PR #25, opened 2015-03-03
[TASK] Add exceptional case for field type group

cedricziel:
Extbase needs the foreign_field option to be set in TCA on field
type group. Otherwise the persistence cannot resolve the corresponding table.


fsuter:
Done in fa273608d110bc9e4a0c90f3a77ec2f6d975f079

PR #26, opened 2015-03-03
[TASK] Add exceptional case for field type group

cedricziel:
Extbase needs the foreign_field option to be set in TCA on field
type group. Otherwise the persistence cannot resolve the corresponding table.


cedricziel:
This is the same commit as in #25 , just cherry-picked for ``latest``

fsuter:
I modified your phrasing. Somehow the merge did not happen. So the commit is ff5049a16a0d1610ad170b4824dc3ec940eb011a.

Actually the best solution would be for Extbase to be fixed, but I agree that this must be mentioned here for now.

PR #31, opened 2015-04-07
[TASK] Remove outdated info about the need of prefixing class names

mostwanted1976:


mostwanted1976:
I just successfully used a class' method without prefixing the class name with "tx_" or "user_" in TYPO3 CMS 6.2.9.

marble:
Thank you. It looks like that this restriction is gone since TYPO3 v6.0, see https://review.typo3.org/#/c/12976/

PR #33, opened 2015-05-07
Update Index.rst

simonschaufi:
removed space in file path

### DIFF
--- a/Documentation/Reference/Columns/Input/Index.rst
+++ b/Documentation/Reference/Columns/Input/Index.rst
@@ -348,7 +348,6 @@ eval
                             return $value . ' [added by PHP]';
                     }
             }
-            ?>
 
          ext_localconf.php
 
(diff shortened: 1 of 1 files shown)


marble:
You're doing good, thank you. You CAN do better, if you provide the patch for the "latest" branch too right away. The we'd only have to click on "Merge". Anyway, thanks for contributing!

fsuter:
Martin, we had a lot of thee in the last few months. This is mostly due to the fact that the stable branch on docs.typo3.org is still "6.2" for all Core manuals. It should be changed to "7" IMO.

simonschaufi:
ups, was without intention.

marble:
Am 12.05.2015 um 12:16 schrieb François Suter:
>
> Martin, we had a lot of thee in the last few months. This is mostly
> due to the fact that the stable branch on docs.typo3.org is still
> "6.2" for all Core manuals. It should be changed to "7" IMO.
>

Oh ha, so you mean, people may mean branch "latest = 7" when they are
actually sending pull requests for 6.2? I didn't verify that.

Offically we are for LTS at 6.2. So that's what "stable" suggests, IMHO.

Having "latest" point to "7" which is actually "7-dev" is misleading, I
think.


Martin

fsuter:
I mean that when people access any official manual on docs.typo3.org, they land by default on "stable", which corresponds to 6.2. Thus they probably don't pay attention and fork 6.2 instead of 7.

That's why I think it would be easier for us if "stable" were pointing to 7 instead of 6.2.

What constitutes "stable" is obviously a matter of dispute. The fact is that the 7 branch has releases which mark it as usable and thus stable, although not as stable as 6.2 LTS does.

I would do the switch, but that's just my opinion.

simonschaufi:
ok thats why i worked on 6.2 instead of master.

simonschaufi:
can somebody take care that my patch is also applied to master or do i have to do it myself?

marble:
I applied it additionally to master already. That's why it took so long; I had been ill and in bed.

marble:
Ok, I got it. And yes, fsuter, that makes the valid point: There ARE stable 7.x versions.
DONE: So I just switched the official manual to "stable" pointing to "latest" - which is "7".

fsuter:
Great, Martin. Thanks.

PR #34, opened 2015-05-20
[TASK] Warn about TCA overrides with dynamicConfigFile

mbrodala:
TCA Overrides are not possible with the old TCA setup via dynamicConfigFile in ext_tables.php,


mbrodala:
Thanks.

PR #35, opened 2015-07-19
[TASK] Add dbType field documentation

alexanderschnitzler:


marble:
Hi Alexander, thanks a lot. With the next commit I added some more improvements to this. If you find more: Just do a pull request. That's great!

PR #36, opened 2015-07-30
[BUGFIX] Mention one table limit of foreign_table

mbrodala:
Extbase expects "foreign_table" to contain exactly one table name compared to "allowed" which can hold a list of table names.


mbrodala:
Thanks.

PR #46, opened 2015-10-29
[TASK] Update `select` renderType and maxitems documentation

yabawock:
Remove the deprecated renderMode from the TCA select documentation and
replace it with the mandatory renderType property. Add a note that
maxitems property is ignored if the renderType is `selectSingle`.

Related: #58521
Related: #69822

### DIFF
--- a/Documentation/Reference/Columns/Select/Index.rst
+++ b/Documentation/Reference/Columns/Select/Index.rst
@@ -1060,50 +1060,31 @@ itemListStyle
 
 .. _columns-select-properties-rendermode:
 
-renderMode
+renderType
 ~~~~~~~~~~
 
 .. container:: table-row
 
    Key
-         renderMode
+         renderType
 
    Datatype
          string (any of keywords)
 
    Description
-         (Only for maxitems > 1)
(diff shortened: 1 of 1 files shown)


marble:
Morton (thinking of "Jabberwocky" :), I like your Twitter avatar! Thanks a lot for the energy to write this reST stuff. 

Some hints:

I've added .. default-role:: code  to Includes.txt. That makes it easier. Instead of :code:`1+1` you only need to write `1+1`.

When you are IN THE SAME FILE you can link to the headline "this is renderType" by just writing `this is renderType`_
Sphinx referencing in the form of :ref:`target`is required as soon as you link to a different file. But it does work in the same file as well, if you prefer that.

The search now DOES work very well. Please search vor 'todo':
 and it will definitely find ALL PAGES that contain 'todo':


Afterwards go to that page and find the hits highlighted:


I have added that "todo" as I noticed this reference to the deprecated "renderMode".

You did a good job. Would you be so kind to make it complete?

Please search for 'renderMode'. You'll see that there still are some hits:


This WORKING SEARCH is a real highlight of the new TYPO3 Sphinx theme!

PR #49, opened 2015-11-12
[FIX] Inline element with foreign_table only has maxitems 1 as default

patkar:
`maxitems` is defaulted to 1, trough `checkValue_checkMax` here: https://github.com/TYPO3/TYPO3.CMS/blob/TYPO3_6-2-15/typo3%2Fsysext%2Fcore%2FClasses%2FDataHandling%2FDataHandler.php#L2927

### DIFF
--- a/Documentation/Reference/Columns/Inline/Index.rst
+++ b/Documentation/Reference/Columns/Inline/Index.rst
@@ -820,6 +820,11 @@ maxitems
          that this is different from types "select" and "group" which default
          to 1.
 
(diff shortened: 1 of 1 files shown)


marble:
Sounds as if you know what you're writing! Thank you very much.

*Q:* Should this go into branch "latest" as well? Can you have a look?

patkar:
I can look, but not at the moment :)

marble:
+1 Take your time!

PR #50, opened 2015-11-18
[TASK] Backport foreign_table appendix from master

mbrodala:
Within "foreign_table" only a single table name is allowed while "allowed" can hold a list of table names.


marble:
:-) Gimme more!

PR #51, opened 2015-11-28
Document ``disableAgeDisplay`` option

PowerKiKi:
Previous documentation was a copy/paste of ``range`` option which was clearly a mistake.
Description is directly taken from Changelog notes (Feature 28243)

### DIFF
--- a/Documentation/Reference/Columns/Input/Index.rst
+++ b/Documentation/Reference/Columns/Input/Index.rst
@@ -530,48 +530,12 @@ disableAgeDisplay
          boolean
 
    Description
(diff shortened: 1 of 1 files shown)


marble:
Perfect, thanks a lot! I took the chance to rework the markup to follow latest insights on how to write reST code. Should be nice looking now. On tablet and Smartphone as well. See https://docs.typo3.org/typo3cms/TCAReference/Reference/Columns/Input/Index.html
*Refresh browser!*

PowerKiKi:
Indeed that's better when text is not leaking across all the layout, thanks for that ! :)

PR #52, opened 2015-12-01
[TASK] Point out the scope of foreign_sortby and symmetric_sortby

ohader:
The disposal of the properties foreign_sortby and symmetric_sortby
are only available for persisting the parent reference with the child,
thus using the property foreign_field. In a comma-separated-value
approach this does not work.


wmdbMattes:
PR looks good

marble:
Perfect, thanks! Is that something for 6.2 as well?

marble:
To reward you and me I've reworked and reformatted the reST code. And it looks nice with the new theme: https://docs.typo3.org/typo3cms/TCAReference/Reference/Columns/Inline/

PR #53, opened 2016-01-07
added required renderType to examples

torvitas:


marble:
Thank you. Send more if you can! I'm trusting you that the fix is necessary and correct.

PR #56, opened 2016-02-04
[BUGFIX] Document default max length for input

mbrodala:
The default value for max is hardcoded to 256 which should be mentioned.


marble:
Do we need that for branch 'latest' as well?

mbrodala:
No, I've checked the 7.6 and master branch and this default limit is not present anymore. It must have been removed with the FormEngine refactor.

marble:
Great. Thanks for checking!

PR #64, opened 2016-07-07
Fix wrong TCA example (common properties > search)

ArminVieweg:
The common property "search" has to be located **inside** of the config-array. It's not a sibling like the example showed before.


marble:
- I updated reST a bit
- applied to 'latest', '7-6', '6-2'.

PR #73, opened 2016-08-30
[TASK] Slightly improve "items" value description

mbrodala:
Ensure values are on 1st level and remarks for these on 2nd level.

### DIFF
--- a/Documentation/Reference/Columns/Select/Index.rst
+++ b/Documentation/Reference/Columns/Select/Index.rst
@@ -135,12 +135,12 @@ items
 
          - Second value is the  **value of the item** .
 
-         - The special value `--div--` is used to insert a non-selectable value
-           that appears as a divider label in the selector box (only for maxitems
-           <=1)
+         	- The special value `--div--` is used to insert a non-selectable value
+           	  that appears as a divider label in the selector box (only for maxitems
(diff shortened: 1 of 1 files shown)


marble:
Hi Mathias, I changed your tabs to three blanks each. And picked a different char to specify the sublist. You can pick any of these characters at any level you want, see [1] "*", "+", "-", "•", "‣", or "⁃", .

Moreover I've indented the whole paragraph "Description" as I would do it: use THREE blanks instead of ONE tab. Do the same for lists, so you can use the tab-key in PhpStorm. Why THREE? Fits best, most beautiful, makes code examples well readable but not tooooo space consuming, and, if that helps, it's in line with the Apple style guide ;-))

[1] http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#bullet-lists

mbrodala:
@marble Thanks for fixing. All I did was pressing `[Tab]` here on Github. ;-)

marble:
Fine! If you wonder about my longy answer: I simply can't help it, I like to let people know :-) And I noticed, you often contribute to documentation. Great!

mbrodala:
I'm always grateful for hints how to improve contributions. :-)

PR #75, opened 2016-09-01
Update index.rst with hint for filter attribute

Blackeye:
As the filter attribute doesn't work in combination with a wizard, I've added a note in the filter section to be aware of this reasons for exclusion.

### DIFF
--- a/Documentation/Reference/Columns/Group/Index.rst
+++ b/Documentation/Reference/Columns/Group/Index.rst
@@ -253,6 +253,10 @@ filter
 
          Define filters for item values.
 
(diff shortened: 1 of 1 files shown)


Documentation/Reference/Columns/Group/Index.rst
```diff
@@ -253,6 +253,10 @@ filter
 
          Define filters for item values.
 
+	.. note::
+
+            Does't work in combination with a wizard.
```
lolli42: Does_n_'t work ;)

PR #76, opened 2016-09-02
Update example for namespaced evaluation classes

sgrossberndt:
Provides an example evaluation using namespaces


marble:
upps, look like I made a mistake with the Git history. But it should be merged just as you sent it. Except: I changed ONE tab into THREE blanks: -> less error prone, much nicer in the frontend, where one tab eats up 8 spaces in a smartphone for example.

sgrossberndt:
Thanks, for the code I just pasted it not paying attention to spaces/tabs. Thanks for the update and merge. Will this be updated in "stable" == 7.6 too?

marble:
Not automatically. Can you send a pull request?

PR #85, opened 2016-11-14
Remove not used versioning_followPages and change versioningWS

tmotyl:
chanegs following https://review.typo3.org/#/c/50462/


tmotyl:
It should be removed from v7 branch as well

marble:
:-))

PR #88, opened 2016-11-21
TCA Manipulations which can't be cached.

TCahn:
It's not possible to put TCA-Manipulation-Code which isn't static (like Page-TS-config specific manipulations) in "Configuration/TCA/<tablename>.php" or "Configuration/TCA/Overrides/<tablename>.php".
You have to put it in ext_tables.php.
Is there a other way?

### DIFF
--- a/Documentation/ExtendingTca/StoringChanges/Index.rst
+++ b/Documentation/ExtendingTca/StoringChanges/Index.rst
@@ -91,3 +91,4 @@ It is also possible to perform some special manipulations on
 :php:`$GLOBALS['TCA']` right before it is stored into cache, thanks to the
 :code:`tcaIsBeingBuilt` signal. This signal was introduced in
 TYPO3 CMS 6.2.1.
+Put all manipulations which can not be cached (for example Page-TS-config specific or website specific TCA-manipulations) into :file:`ext_tables.php`.


liayn:
IMO pageTS config is not related to TCA here.

ohader:
The section which shall be modified is about using the hook `tcaIsBeingBuilt` to modify TCA prior to be cached. The suggested way to override TCA is to use the `Configuration/TCA/Overrides` folder. Defining TCA in `ext_tables.php` is not suggested and also might not work anymore in future versions of TYPO3 CMS.

marble:
I'm not sure what to do with this PR. Shouln't it become something like: 
""".. tip:: (green box)
Put all manipulations which can not be cached (for example Page-TS-config specific or website specific TCA-manipulations) into :file:`ext_tables.php`.""" Maybe with a link to Tsconfig manual?

fsuter:
I understand the need for having a list of DOs and DONTs for TCA/Overrides files. I think that's what @TCahn wants to say here. It's not really about TSConfig, it's about what you may put in TCA/Overrides or not. What makes this not as clear as one may think, is that it is okay to put API calls in such files (typically to method of ExtensionManagementUtility) as long as those calls modify the TCA under the hood, which you can know only by looking at the source code.

So maybe that's something that could be added.

And anyway, as has been mentioned already, such information should not be added at the bottom of the page, but in the "Storing in the Overrides folder" section.

marble:
I'm not expert enough to handle the PR. I suggest somebody else cares.

fsuter:
I could handle it, but I would gladly have a Core Team member's input on the dos and donts.

TCahn:
Thank you to all! :-)

When you have the following:
-	A TYPO3 Installation with 2 Websites "Domain 1" & "Domain 2".
-	A Extension "News" (for example) with totally different TCA Configurations for "Domain 1" and "Domain 2" …and with totally different BE Forms, with totally different Language-Files, with different Icons….
You cannot put the different TCAs in "/TCA/xyz.php" or "Configuration/TCA/Overrides/xyz.php", because all TCA-Modifications are cached and for the complete Typo3 Installation.
**Where can you put the TCA-Modifications for Website "Domain 1" & "Domain 2"???**
(I put the modifications of the cached "Basic TCA" into ext_tables.php – so this modifications are not cached and everything is fine and "Domain 1" & "Domain 2" are totally different. Wonderful! :-)
I used: "$currentPageTSconfig = \TYPO3\CMS\Backend\Utility\BackendUtility::getPagesTSconfig($currentBEPageID);" to get the PageTSconfig for "Domain 1" & for "Domain 2")

@ohader: You wrote: "Defining TCA in ext_tables.php is not suggested and also might not work anymore in future versions of TYPO3 CMS."
=> Where can I put the TCA-Modifications when this is not working in future versions of TYPO3 CMS?

liayn:
@TCahn If you modify the FormEngine (so you change the configuration gathered from TCA) with TSconfig, this can still go into ext_tables.php. But you do NOT actually change the TCA, this is where the misunderstanding comes from.

TCA must be defined in `Configuration/TCA/tablename.php` and may be overridden in `Configuration/TCA/Overrides/tablename.php`. The combination of those definitions (the resulting TCA) is then fully cached.

In the described usecase I would rather argue that the used tools do not match the use case. If really TCA must be so much different for two cases, I would propose to make it 2 extensions, each tailored for the actual usecase.

fsuter:
I will close this pull request, as the question has been answered IMO.

As for the proposed improvement, I have made an alternate version in https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/commit/661033a208a26aab7a1ecce5bc78a36ffb50901f

PR #93, opened 2017-02-01
[TASK] Replace $TCA with $GLOBALS['TCA']

MonaMuzaffar:


marble:
Thank you. Great you took the time to send this PR. Send more! FYI: I tried to be consistent and marked the code as :php:`...`. In the headline the simple $TCA gives a better understanding, I think. Note: If you make the headline longer, the line below needs at least to be the same as long. See http://docutils.sourceforge.net/docs/user/rst/quickstart.html#sections

PR #94, opened 2017-02-02
Add maxItemsInResultList for SuggestWizard

sgrossberndt:
Add the missing documentation of maxItemsInResultList, which has always been available.
Needs to be backported to 7.6 and 6.2.


sgrossberndt:
Patch for latest, added pull requests for 7.6 and 6.2 too.

PR #136, opened 2017-04-05
[Bugfix] Fix typo in Userfunc example

Smove:


marble:
Just be reading: You are turning correct code into wrong code. Did you test your patch?

ohader:
The change is correct, the single-quote in the beginning is superfluous and is correctly removed by this pull-request

PR #138, opened 2017-04-20
Update custom configuration for FAL

manuelselbach:
To add custom configuration for the file abstraction layer the overrideChildTca has to be used instead of foreign_types.


lolli42:
that's correct. thank you manuel!

PR #143, opened 2017-06-07
Improved example of foreign_table_where

Eddcapone:
It was unclear, i thought you need the [ ] cling squares and my code failed with errors beacuse of this.

### DIFF
--- a/Documentation/Reference/Columns/Select/Index.rst
+++ b/Documentation/Reference/Columns/Select/Index.rst
@@ -1,4 +1,4 @@
-﻿.. ==================================================
+.. ==================================================
 .. FOR YOUR INFORMATION
 .. --------------------------------------------------
 .. -*- coding: utf-8 -*- with BOM.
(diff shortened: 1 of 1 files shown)


DavidBruchmann:
I think the old version is better even you had problems because in general it can be any table including fe_users, be_users, pages, tt_content, etc. So the notation tx_myextension_domain_model_mytable might be more confusing than helping.

Eddcapone:
Thats true, then whe should just use **table** instead of **tx_myextension_domain_model_mytable** ? It is very confusing with the [] because everyone things that they are needed.

mbrodala:
Personally I'm in favor of changing this but I'm open as to how. But I agree that naming a placeholder (`[foreign_table]`) exactly like an option (`foreign_table`) can lead to confusion.

If you are worried that people might not understand that they can use arbitrary tables, I'd suggest one example using e.g. `pages` or `tt_content` and another one with something like `tx_myext_mytable`.

lolli42:
What about aTable or aTableName (without brackets)?

pi-phi:
1st If you want to use something like EBNF you need to inform readers about your used syntax.
2nd you need to use the same notification all the time to avoid misunderstandings (like the origin here)

if you use no continous syntax, you need to explain your intention with good examples, where everyone recognizes the main parts, which is always helpful as this might also be found be search engines.

DavidBruchmann:
In general there should be a syntax-guide in the beginning of the document to describe what kind of syntax is used for which cases in the document (or all documents). This guide can be a reference to another document. Then this one (or all) documents have to use this syntax and every user should be aware of the syntax or use the (optional referenced) guide.
This also assures that the syntax is the same in the document(s).
Apart from that, learning that brackets mark placeholders might be hard for newbies but it's quite common.
One Improvement could be to write inside the brackets only in upper case, but that depends on the rules in the syntax-guide and it's also possible to avoid upper case to make clear that table-names are written in lower case.
Brackets are usually a good method to mark placeholders but like before it depends on the syntax-rules.
So in my opinion everything is open, but like @pi-phi said it should be constant in the whole document(s) and follow the same rules (of the syntax-guide).
Using several examples like @mbrodala said is a good method to include different use-cases and make reading in the whole more intuitive and understanding easier.

lolli42:
Could we keep this change straight? Using something like aTableName or a_table_name should be pretty clear this is a placeholder and how a specific one should look like, that's obvious on reading without further documents.

marble:
Cool to see that much interest :-)) Should be pretty clear now https://docs.typo3.org/typo3cms/TCAReference/6.2/Reference/Columns/Select/Index.html#foreign-table-where

DavidBruchmann:
Hehe, sometimes the effort of professionals trying to think how beginners or non-professionals (could) think ;-)

marble:
@DavidBruchmann I don't understand your sentence ...

PR #144, opened 2017-06-13
Fix `suggestOptions` code example

sgrossberndt:
The code example for `suggestOptions` still contained the old code for wizards[suggest]


lolli42:
Uh, that slipped through. Thank you!

PR #145, opened 2017-06-14
[BUGFIX] Adjust description for localizeReferencesAtParentLocalization

ohader:
'localizeReferencesAtParentLocalization' can be used for MM-tables in TCA
types 'select' and 'group'. The remark on TCA type 'inline' is not valid,
since it has a distince 'localizeChildrenAtParentLocalization' setting.

Relates: https://forge.typo3.org/issues/81562


ohader:
The description is wrong in docs for older TYPO3 versions as well...

marble:
The structure of the older branches is quite different, so cherrypicking isn't possible. I add an issue that the older docs still need to be changed.

PR #158, opened 2017-09-15
Missing crucial renderType in example

groso:
Adds crucial renderType to selectMultipleSideBySide example


marble:
Thank you. Send more!

PR #166, opened 2018-01-16
add position where to find the "pi_flexform"

franzholz:
And the selection of "$GLOBALS['TCA']" is required.


marble:
Tip: It would be clever to use "code" markup like so: `pi_flexform`. Looks nice and the underscore need not be escaped.

PR #167, opened 2018-01-24
[BUGFIX] Fix level of sections in introduction

mbrodala:
The sections "Properties scope" and "Glossary" are not related to "TCA main array structure".


marble:
very good

PR #172, opened 2018-02-20
[BUGFIX] Fix name of table 'sys_file_metadata'

frauliessmann:


marble:
Thank you - send more! Nu hatDatDda one error less :-))

PR #180, opened 2018-04-09
Document new renderTypes and fix some other issues

sypets:
- Fix some errors with error section levels
- Document new renderTypes
- Fix orphaned AllowLanguageSynchronization

Fixes: #181
Fixes: #179
Fixes: #178
Fixes: #176

PR #182, opened 2018-04-11
TYPO3 8.7: Fix section level issues and more

sypets:
This is the same as [PR 180](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/180) for TYPO3 8.7 branch (without new renderTypes)


marble:
@sypets (1) see commit messages (2) I didn't check: Do we have the same issues in 'latest'? Should we start there?

marble:
Hhm, too bad, maybe I should have looked at https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/180 first. But I didn't. Do my commits here still make sense?

sypets:
Martin, I think we should talk directly on Slack. 

1. If I understood you correctly, you are saying, all included files should be .txt, not .rst? 

The TCA reference includes a large number of files. Almost every file in Documentation/ColumnsConfig/Properties is included. They are all .rst. When I make changes I usually try to conform to what is already being done. I am not aware of all the best practices yet. 

Also, if your editor or IDE is configured for .rst, the editing rules for reST will not apply for .txt files. 

I think it would make sense to more clearly mark files that will be included though. Maybe a subdirectory `Include` or something like that. 

2. Why do you remove Documentation-GENERATED-temp from .gitignore? That is the path that is being generated. I think it is very convenient to have it in the .gitignore by default. 

Also I would suggest to use the same .gitignore in all repositories here. That could save everyone some time. 

3. About the section hierarchy: That's what I am trying to fix in the first place. It was wrong and it is not working in the currently rendered version: e.g. in type check the other properties should not be under renderType, they should be on the same level: https://docs.typo3.org/typo3cms/TCAReference/ColumnsConfig/Type/Check.html#rendertype-default

Also, some subheadings are currently missing completely: Everything after fieldWizard: https://docs.typo3.org/typo3cms/TCAReference/ColumnsConfig/Type/Check.html#fieldwizard

see

```
../../TheProject/Documentation/ColumnsConfig/FieldWizard/DefaultLanguageDifferences.rst:1: SEVERE: Title level inconsistent:

defaultLanguageDifferences
^^^^^^^^^^^^^^^^^^^^^^^^^^
../../TheProject/Documentation/ColumnsConfig/FieldWizard/LocalizationStateSelector.rst:1: SEVERE: Title level inconsistent:

localizationStateSelector
^^^^^^^^^^^^^^^^^^^^^^^^^
../../TheProject/Documentation/ColumnsConfig/FieldWizard/OtherLanguageContent.rst:1: SEVERE: Title level inconsistent:
```
https://docs.typo3.org/typo3cms/TCAReference/_buildinfo/warnings.txt

So, _I hope_ I got it right, I tested the rendering. I don't think I messed it up any more than it is now. 

4. Please clarify what you mean by "should have a proper doctitle markup".

marble:
- Everything is fine, you're doing a great job. 
- Just be patient for a moment and don't be surprise when I add notes here in the comments piece by piece. 
- Once done I let you know and we'll do a hangout.

1. The .gitignore isn't thought well through, it simply had been there for some reason. But it does what you want. It has the line `*GENERATED*`. Actually I have the following in MY global .gitignore file on my machine. And we can still think about what the minimum .gitignore file should contain. And compared to what was there before it's simply ASCII sorted plus `*GENERATED*` (which would be the only line we really need and `*NOT_VERSIONED*` (which I'll remove from the PR):
```➜  ~ cat ~/.gitignore_global 
*GENERATED*
*NOT_VERSIONED*
.env
.idea
.volumes
```

marble:
@sypets 
1. With this version of the PR there are no warnings left.
2. I would check this version in. Do you want to render it locally first?
3. I've renamed the included files to `*.rst.txt`. You can set a rule in PhpStorm to detect those files as reST files.

PR #187, opened 2018-05-18
[TASK] Mention MySQL strict mode

mbrodala:


marble:
I hope you agree with "may" instead of "will". If not, recurse().

PR #193, opened 2018-08-10
Remove duplicate text block

koehnlein:
The removed and the included text are nearly the same. There are tiny differences ("whether" vs. "if"). So if you think I removed the wrong of them, feel free to reject my pull request and change it the other way.


sypets:
The included file contains another header for allowLanguageSynchronization so you would have 2 headers then. If you remove the header in AllowLanguageSynchronization.rst, the heading levels would be wrong. 

I would have to check, if we can / should remove the header in the include file or if it is currently really needed in 2 places.

PR #202, opened 2018-09-13
Remove the mention to the parent object

janhelke:
The code shows, that the parent object is no longer passed to the user function.See \TYPO3\CMS\Core\Utility\GeneralUtility::2195 as example. So there is no need to mention it in this documentation.


marble:
Trusting you. Thanke you. Send more!

PR #206, opened 2018-11-29
Add warning to use input/text fields for label

manuelselbach:
Please find a more detailed description at the following issue on forge:
https://forge.typo3.org/issues/87041


sypets:
@manuelselbach Thanks for doing this. 

Just to be very clear here: You will get problems if you use any field for label which is not of type 'label' or 'text'? Then, additionally pointing out that 'CType' is a problem, is redundant, IMHO.

I think this is mergable, but I think we should make it _even more clear_, what you can and shouldn't do.

sypets:
@manuelselbach Thanks for reacting so quickly!

PR #208, opened 2018-12-12
Add generatorOption "replacements" to example

ochorocho:
There is a new option available for slugs in TYPO3 9.5.2 which is calle `replacements`.


ayacoo:
Maybe you can add it under "generatorOptions" description, too?

PR #212, opened 2019-01-15
Correct class names with namespace

sypets:


sypets:
I believe, in PHP the fully qualified path for class name should be with with leading \, no double \. See also https://docs.typo3.org/typo3cms/TCAReference/ColumnsConfig/Type/User.html#examples

DanielSiepmann:
You're right @sypets. Only inside of single quoted strings you will see them double, and even there it's not always required.

PR #217, opened 2019-02-11
Update LinkPopup.rst.txt

maikusch:
Added an example, because this special config has no example in the docs.


DanielSiepmann:
Thanks for providing an working example to make it easier to use.

PR #230, opened 2019-08-20
Update Slug.rst

rpdmind:
Document possibility to use multiple fields as fallback for each other


DanielSiepmann:
There is a small difference. The array contains only a single string with comma separated values.

I'm not into this topic, but this might result in a different behaviour? @susannemoog, can you check that once more?

DanielSiepmann:
No reaction since last comment.

There were also some improvements already to polish slug generation documentation.

I'm closing for this reasons.

PR #232, opened 2019-09-05
<sheets> instead of <sheet>

franzholz:
See example:

https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/DataFormats/T3datastructure/SheetReferences/Index.html

### DIFF
--- a/Documentation/ColumnsConfig/Type/Flex.rst
+++ b/Documentation/ColumnsConfig/Type/Flex.rst
@@ -389,7 +389,7 @@ format. The structure is as follows:
 .. _columns-flex-data-format-array-sheet:
 
 <sheets>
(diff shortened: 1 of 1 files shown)


Documentation/ColumnsConfig/Type/Flex.rst
```diff
@@ -388,11 +388,11 @@ format. The structure is as follows:
 
 .. _columns-flex-data-format-array-sheet:
 
-<sheet>
+<sheets>
 ^^^^^^^
```
DanielSiepmann: The "underline" of the headline has to be as long as the headline itself. In this case the headline was increased by one character, so underline has to be increased also by one.

PR #234, opened 2019-09-11
Update slug generatorOptions fields examples

h3nn3s:
This updates tries to provide some more specific examples due to a not accurate documentation for me.

To generate a slug with "take this field if not empty, otherwise take that one (TS: nav_title // title)" you have to use a nested array. If you take a simple array, the two fields are concatenated instead of choosing only one field. The examples no. 2 and 3 point this out.


sypets:
See also related issue on forge: https://forge.typo3.org/issues/89390

Related:

* forge: https://forge.typo3.org/issues/89390
* GitHub: #244

Must also resolve conflicts.

DanielSiepmann:
Also merged into 9.5.

PR #241, opened 2019-10-04
Change slug type description

kevinmeckl:
The specification of the generator option fields is not correct. This has been noticed during testing. The shortcuts are interpretted as an array.


sypets:
Are you sure about this?

IMHO `field` should be an array (and not a comma delimited string), as currently documented: 

```
'fields' => ['nav_title', 'title']
https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Slug.html#generatoroptions

```
You can also see this in the working test class:

https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Tests/Unit/DataHandling/SlugHelperTest.php

If this does not work as currently documented, maybe you should open a bug report: https://forge.typo3.org/projects/typo3cms-core/issues

P.S. Please feel free to reopen and clarify if I made a mistake here.

sypets:
It seems there is in fact an error in the documentation. Didn't realize that at the time. See open issue: https://forge.typo3.org/issues/89390

Also there is (at least) one more PR:

* PR: #234 
* see also GitHub issue: #244

PR #243, opened 2019-10-10
Update SlugGeneratorOptions.rst.txt

ayacoo:


Documentation/ColumnsConfig/Properties/SlugGeneratorOptions.rst.txt
```diff
@@ -49,7 +47,7 @@ generatorOptions
          'config' => [
              'type' => 'slug',
              'generatorOptions' => [
-                 'fields' => ['title'],
+                 'fields' => [['title']],
```
DanielSiepmann: I thought this example was also wrong and adjusted it. Is this correct @ayacoo ?

DanielSiepmann: As discussed via Slack, this is kept as it was before. The information that either an array with strings, or an sub array can be used, is now an info box, to raise awareness.

PR #250, opened 2019-11-20
Fix SuggestWizardDefaultReceiver class path

georg-tiefenbrunn:
`TYPO3\\CMS\\Backend\\Form\\Wizard\\SuggestWizardDefaultReceiver` instead of `TYPO3\\CMS\\Backend\\Form\\WizardSuggestWizardDefaultReceiver`.

### DIFF
--- a/Documentation/ColumnsConfig/Properties/GroupSuggestOptions.rst.txt
+++ b/Documentation/ColumnsConfig/Properties/GroupSuggestOptions.rst.txt
@@ -100,7 +100,7 @@ suggestOptions
           ],
 
     receiverClass (string)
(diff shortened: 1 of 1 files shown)


sypets:
In the future, it is not necessary to create a PR for all version branches. We can take care of that when merging (as in the TYPO3 core). 

It is helpful though to specify the revisions for which your patch applies. If you already researched that, we don't have to. 

You can do that by adding for example the following line to your commit message:

```Releases: master, 9.5```

see https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CommitMessages.html

(We are using the same conventions that are used in the TYPO3 core for that to make it easier for core developers to contribute to docs).


When there is only one commit in the PR, the PR will automatically contain the text from the commit message. 


Thank you for your contributions :thumbsup:

PR #251, opened 2019-11-25
Rearrange examples on checkbox page

sypets:
Before, all the screenshots (how it looks) were displayed first
and then all the code examples.

This change displays the source snippet directly under the screenshot.

Additionally:

- add 2 examples for FlexForms (using the exact same functionality as the TCA in PHP)
- add a 1 line description pointing out what differentiates each
  example from the others
- add headlines as separators


DanielSiepmann:
I like that change. Would be cool to have that for all types :)

sypets:
> I like that change. Would be cool to have that for all types :)

Added as issue #253

PR #254, opened 2020-01-06
Update LinkPopup.rst.txt

TheNaderio:
Added `folder` to the "Possible values are...." list for blindLinkOptions.


mbrodala:
Thanks for the fix!

PR #266, opened 2020-03-08
Add eval null to make it work for mysql strict mode

georgringer:
see https://forge.typo3.org/issues/88712


sypets:
Cherry-picked to 9.5

Issue #258, opened 2020-01-23
Overview of possibilities for renderType since 9.5 missing?


sypets:
Version 7.6 of TCA reference has an overview of all possible renderTypes:

![renderTypes](https://user-images.githubusercontent.com/13206455/72981977-5406fe80-3dde-11ea-967e-dc3f9750a300.png)

https://docs.typo3.org/m/typo3/reference-tca/7.6/en-us/Reference/Columns/Select/Index.html#rendertype

This seems to be misssing for later versions (or is more difficult to find), e.g. see 9.5: https://docs.typo3.org/m/typo3/reference-tca/9.5/en-us/ColumnsConfig/Type/Select.html

Issue #253, opened 2019-12-20
Change order of examples to put code-block directly under each screenshot


sypets:
in PR #251 some examples were rearranged in the following manner:

> before there were several screenshots in one block and then several code examples on the "checkbox" page. The PR split that up into several examples. That way you could see the screenshot (how it looks) directly above the code snippet

I have not looked at the other examples. Possibly, we could check them if there are other cases where the examples could be improved and do it consistently. 

---

sypets:
1. Example where examples were rearranged (to show screenshot directly above the code snippet):

![example1](https://user-images.githubusercontent.com/13206455/73198874-1da9e600-4134-11ea-82ee-89f898e78af0.png)


https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Check.html

2. Example where examples are in one list of screenshots and then one list of code snippets



https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Select.html

sypets:
Was resolved with #281 and

PR #281, opened 2020-05-07
Split up types with several renderTypes into subpages

sypets:
The select type can have 5 different renderTypes. For each renderType a
number of properties exist and are listed.
    
This used to be all on one page which made it difficult to determine
which properties exist for one rendertype and when scrolling it could
easily happen that you were looking at a property for a different renderType.
    
This patch makes the following changes:
    
- Put properties and examples for each renderType on a subpage
- Add a table of contents of all properties at the beginning of the
   properties section
- Change the ordering in the examples to list the code directly under
   the screenshot and not all screenshots and then all codeblocks (as already
  done with the checkbox page).
    
Resolves: #258
Related: #253

Releases: master, 10.4, 9.5


Documentation/ColumnsConfig/Type/Select.rst
```diff
@@ -31,617 +32,28 @@ In case of the n:m (e.g. Organization has multiple categories) the uid of the se
...
-.. include:: ../Properties/CommonSize.rst.txt
-
-.. _columns-select-properties-treeconfig:
-.. include:: ../Properties/SelectTreeConfig.rst.txt
+The chosen renderType influences the behaviour of the element and how it will
+be displayed.
+
+The following renderTypes are available:
+
+* :ref:`selectSingle <columns-select-rendertype-selectSingle>`: Select one
```
DanielSiepmann, line 40: I prefer definition lists, what do you think?

PR #282, opened 2020-05-07
Change order of examples in select type

sypets:
Put each code block directly under the screenshot as already done for the
'check' type.


sypets:
Is now done in other PR #281

PR #291, opened 2020-06-23
Update CtrlCopyAfterDuplFields.rst.txt

IndyIndyIndy:
Moving a record won't apply the effect of "copyAfterDuplFields" since TYPO3 9.5.


sypets:
Backported to master, 10.4, 9.5

PR #301, opened 2020-09-23
[DOC] Remove listing of unsupported option showIconTable

NamelessCoder:
This option is no longer supported on 10.4 and listing it is misleading.


sypets:
Releases: master, 10.4, 9.5, 8.6

showIconTable was introduced in 7.6 and [removed in 8.6](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.6/Deprecation-79440-TcaChanges.html).

PR #303, opened 2020-09-30
Update InlineForeignLabel.rst.txt

beyring:
The description did not point out clearly that you have to insert the field name of the irre element that you want to have as title here.


susannemoog:
Thank you for your contribution 👍

PR #306, opened 2020-10-04
[TASK] Avoid visual quote style in displayCond section

brotkrueml:
Additionally, fix the indentation to 3 spaces.

### DIFF
--- a/Documentation/Columns/ColumnsDisplayCond.rst.txt
+++ b/Documentation/Columns/ColumnsDisplayCond.rst.txt
@@ -2,159 +2,159 @@ displayCond
 -----------
 
 :aspect:`Datatype`
-    string / array
+   string / array
 
 :aspect:`Scope`
-    Display
+   Display
 
 :aspect:`Description`
-    Contains one or more condition rules for whether to display the field or not.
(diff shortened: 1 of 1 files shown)


sypets:
About the indenting: You may find the manuals are currently not consistently indented. Some are indented with 3 spaces, some with 4. The official style is [3 spaces](https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CodingGuidelines.html). I believe that is also what is in the sphinx example manual and is commonly used but not used everywhere.

About the 3 space there were some objections a while ago. 

Anyways, I would not put too much effort into changing the indenting style now until we have tools to enforce this and automatically convert. 

Apart from that it might be a good idea to find out if there are still objections about the 3-space rules or even if this is still the best choice and get everyone on board using consistent CGL in the docs (like in the core).

sypets:
See also:

* https://github.com/TYPO3-Documentation/T3DocTeam/issues/150
* https://github.com/TYPO3-Documentation/T3DocTeam/issues/57

sypets:
P.S. does **not** mean we should not merge this PR.

PR #313, opened 2020-10-12
[TASK] Add missing code example for group file

### LINKED ISSUE #309: Type "group" example with file relations has no example snippet
All other examples have a snippet

https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Group.html#group-file-relation

runepiper:
I hope my first pull request is ok.

Will resolve #309


Documentation/ColumnsConfig/Type/Group.rst
```diff
@@ -119,6 +119,27 @@ Group file relation
 
    Group file relation with some selected files (group_file_1)
 
+.. code-block:: php
```
mbrodala, line 122: We should drop the whole example instead:

> Types “file” and “file_reference” should not be used anymore. Use FAL references instead.

Also see https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.0/Breaking-87193-DeprecatedFunctionalityRemoved.html

mbrodala (changes requested):

Documentation/ColumnsConfig/Type/Group.rst
sypets: Ah, sorry about that. I created the issue and tagged it with "good first issue". Was not the best choice.

@runepiper Thank you for your pull request. We should do what @mbrodala wrote.
 
You can make additional changes to this pull request or someone else can finish it off (since this is your first PR and I am not sure how familiar you are with the workflow).

Just let us know if you wish to continue here.

runepiper: Just added a commit which removes the whole example.

mbrodala (approved):

Documentation/ColumnsConfig/Type/Group.rst
mbrodala: Thanks. :-)

PR #317, opened 2020-11-04
[TASK] Remove internal_type file related content

susannemoog:
See Core 87193
See https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/849


Documentation/ColumnsConfig/Properties/GroupInternalType.rst.txt
```diff
@@ -25,6 +17,3 @@
       This will create a field where database records can be attached as references.
 
     The default value is none of them - you must specify one of the values correctly!
-
-    .. important::
-        Types "file" and "file\_reference" should not be used anymore. Use FAL references instead.
```
linawolf, line 30: I would leave it in and change the wording: "have been removed" and add some information about possible migration paths for ppl who are updating.

linawolf (commented):
Interestingly I have been on this topic but yesterday. I saw that internal_type=file has been outdated and regretted it a lot. In my use case (choosing background images from a folder) I didnt really find a better solution. Using irre and fal images is no solution One might be to use a select field and a userfunc to produce the items. Not very elegant but seemingly the only solution. However I havent found a better solution. If there is one we should hint to it.

susannemoog:
> 
> 
> Interestingly I have been on this topic but yesterday. I saw that internal_type=file has been outdated and regretted it a lot. In my use case (choosing background images from a folder) I didnt really find a better solution. Using irre and fal images is no solution

Why not? I don't know of a better solution, the changelog entry for that basically says use FAL for everything file related - maybe check directly in typo3-coredev if someone has a better plan.

linawolf:
Irre is kind of not a solution to replace Content Elements such as these. So a Migration path would be great. If I find one I will add it. For the time being i think your PR is fine.

image

linawolf (approved):

mbrodala:
@linawolf In this case you should register your icons with the `IconRegistry` and add a custom `itemsProcFunc` for the `select` field for this. In the frontend you can then use the core `IconViewHelper`. But let's discuss this in Slack if you like. :-)

mbrodala (approved):

DavidBruchmann:
@linawolf the 6th icon with the 3 cookies needs some work :D

PR #323, opened 2020-11-17
Remove false information about RealURl from Slug.rst

dmitryd:
The statement about RealURL is incorrect: path segments there can contain slashes if the whole path is overwritten, just like slugs. Therefore the statement provides false information about RealURL and should be removed.

### DIFF
--- a/Documentation/ColumnsConfig/Type/Slug.rst
+++ b/Documentation/ColumnsConfig/Type/Slug.rst
@@ -23,7 +23,7 @@ With a URL like `https://www.typo3.org/ch/community/values/core-values/` a URL s
 Within TYPO3, a slug is always part of the URL "path" - it does not contain scheme, host, HTTP verb, etc.
 
 A slug is usually added to a TCA-based database table, containing some rules for evaluation and definition.
-Slugs can contain slashes.
+A slug is a segment of a URL, but it is not limited to be separated by slashes. Therefore, a slug can contain slashes.
 
 If a TCA table contains a field called "slug", it needs to be filled for every existing record. It can
 be shown and edited via regular Backend Forms, and is also evaluated during persistence via DataHandler.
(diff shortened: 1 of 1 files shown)


mbrodala (approved):

mbrodala:
Thanks for the fix. I went ahead and restored the long version minus the RealURL part. :-)

sypets:
I up-ported to 10.4 and master.

> A slug is a segment of a URL, but it is not limited to be separated by slashes. Therefore, a slug can contain slashes.

It's already merged but I don't find the phrasing very good - but that was already that way before.  I will try to come up with a suggestion.

I find the removal of the reference to realurl an improvement - can't always assume people used realurl before and as Dimitry pointed out wasn't that accurate.

Documentation/ColumnsConfig/Type/Slug.rst
```diff
@@ -23,9 +23,7 @@ With a URL like `https://www.typo3.org/ch/community/values/core-values/` a URL s
 Within TYPO3, a slug is always part of the URL "path" - it does not contain scheme, host, HTTP verb, etc.
 
 A slug is usually added to a TCA-based database table, containing some rules for evaluation and definition.
-
-In contrast to concepts within RealURL of "URL segments", a slug is a segment of a URL, but it is not limited
-to be separated by slashes. Therefore, a slug can contain slashes.
+A slug is a segment of a URL, but it is not limited to be separated by slashes. Therefore, a slug can contain slashes.
```
DavidBruchmann, line 26: > Therefore, a slug can contain slashes
... in a single path-segment.

sypets:
Please open new PR for further changes.

Documentation/ColumnsConfig/Type/Slug.rst
DavidBruchmann: see #324

dmitryd: @DavidBruchmann I am happy with any text as long as it does not say wrong things about RealURL :)

DavidBruchmann: @dmitryd nice that you can be satisfied so easy :)  
I changed it completely and it's already merged.

PR #328, opened 2020-11-26
Add note about excluded search fields

### LINKED ISSUE #325: dbType 'date' not useable in searchFields
## Problem
Adding a column of `dbType => date` and `renderType => inputDateTime` into the `searchFields` list yields no results when searching for a record with this date in the List module.

## Question
Am I right to assume that `date` fields are being omitted in [Line 310 of LiveSearch.php](https://github.com/TYPO3/TYPO3.CMS/blob/10.4/typo3/sysext/backend/Classes/Search/LiveSearch/LiveSearch.php#L310)?

## Impact
If so, this should definitely be mentioned in the documentation of the `searchFields` section [here](https://docs.typo3.org/m/typo3/reference-tca/10.4/en-us/Ctrl/Index.html#searchfields) and the `search` section of the `inputDateTime` render type [here](https://docs.typo3.org/m/typo3/reference-tca/10.4/en-us/ColumnsConfig/Type/inputDateTime.html#search)

remonator:
Mention fields that are excluded from LiveSearch by default

Closes #325

### DIFF
--- a/Documentation/ColumnsConfig/Type/inputDateTime.rst
+++ b/Documentation/ColumnsConfig/Type/inputDateTime.rst
@@ -143,12 +143,6 @@ readOnly
 search
 ------
 
-.. note::
-
-    Input fields of renderType ``date``, ``time`` and ``int`` are excluded from search by default.
-    To include them, use this hook:
-    :doc:`t3core:Changelog/9.2/Feature-71911-AddConstraintHookInDatabaseRecordListMakeSearchString`.
-
 .. include:: ../Properties/CommonSearch.rst.txt
 
 size
--- a/Documentation/ColumnsConfig/Type/inputDateTime.rst
+++ b/Documentation/ColumnsConfig/Type/inputDateTime.rst
@@ -143,6 +143,12 @@ readOnly
 search
 ------
 
+.. note::
+
+    Input fields of renderType ``date``, ``time`` and ``int`` are excluded from search by default.
+    To include them, use this hook:
+    :doc:`t3core:Changelog/9.2/Feature-71911-AddConstraintHookInDatabaseRecordListMakeSearchString`.
+
 .. include:: ../Properties/CommonSearch.rst.txt
 
 size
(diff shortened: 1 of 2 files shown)


Documentation/ColumnsConfig/Type/inputDateTime.rst
```diff
@@ -143,6 +143,12 @@ readOnly
 search
 ------
 
+.. note::
+
+    Input fields of renderType ``date``, ``time`` and ``int`` are excluded from search by default.
```
mbrodala, line 148: There are no `renderType` options like these by default. Did you mean `eval`? Also this should be mentioned in general for `input` since there is no `renderType` check actually.

Documentation/Ctrl/CtrlSearchFields.rst.txt
```diff
@@ -11,6 +11,8 @@ searchFields
     Comma-separated list of fields from the table that will be included when searching for records in the TYPO3 backend.
     No record from a table will ever be found if that table does not have "searchFields" defined.
 
+    Note that certain fields of type :ref:`input <columns-input>` are excluded by default.
```
mbrodala, line 14: We should move the hint here including the mention of the hook.

mbrodala (changes requested):

Documentation/Ctrl/CtrlSearchFields.rst.txt
remonator: Please check the latest changes, if that's better

Documentation/ColumnsConfig/Type/inputDateTime.rst
remonator: I removed it from this position in favor of the hint in `CtrlSearchFields.rst`

mbrodala (approved):

Documentation/Ctrl/CtrlSearchFields.rst.txt
mbrodala: Yep, sounds good now!

PR #329, opened 2020-11-29
[BUGFIX] Fix case of mentioned defaultAsReadonly value

tbal:
Documentation of l10n_mode mentions l10n_display setting with value 'defaultAsRead**O**nly'. The value of the setting is case sensitive and must be 'defaultAsRead**o**nly'.

This also affects 10.4 and 9.5 branches. Should I create separate Pull Requests for those or is there a merge strategy which allows to apply the patch to multiple branches?


mbrodala (approved):

mbrodala:
Thanks for the fix, I backported it to v10 and v9.

PR #332, opened 2020-12-23
Minor text improvements

sypets:


Documentation/Index.rst
```diff
@@ -48,12 +48,11 @@ at hand.
 
 **Versions**
 
-TCA properties change between major versions, this main array had significant
+TCA properties change between major versions. This manual had significant
 refactorings especially between core versions 6.2 to 7 LTS, and again between 7
```
mbrodala, line 52: Since this part is 1) outdated and 2) true for any LTS we should rephrase it to generally mention possible changes between LTS versions.

linawolf: do you have an idea, how to rephrase it? I was also thinking about the sentence but had no idea.

linawolf (commented):
I would also

Documentation/Index.rst
mbrodala: _... refactorings between each TYPO3 LTS release._

This should be enough.

mbrodala: Also "refactoring" is probably the wrong term since this is mostly meant for actual code. So we should maybe use "changes" or similar instead.

linawolf (approved):

PR #345, opened 2021-01-18
Verbosely explain "uniqueInSite"

mbrodala:

PR #348, opened 2021-02-03
Add unique as eval keyword

froemken:
Add a hint to prevent using uniqueInSite within a multi-domain setup.

### DIFF
--- a/Documentation/ColumnsConfig/Properties/SlugEval.rst.txt
+++ b/Documentation/ColumnsConfig/Properties/SlugEval.rst.txt
@@ -9,8 +9,12 @@
 
     Keywords:
 
+    unique
+        evaluate if a record is unique in its table.
+
     uniqueInSite
-        evaluate if a record is unique in a page tree (specific to a language).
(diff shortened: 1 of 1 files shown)


mbrodala:
Actually this was already done with #345 but the backport to 10.4 and 9.5 is still missing ...

froemken:
So what should I do? Take the description from master to this patch or remove my updates regarding uniqueInSite but keep new keyword "unique"?

mbrodala:
If you are fine with the master text you can copy it here, then we have at least the 10.4 backport covered already.

Also `unique` is documented properly in `master`, so it would make sense backporting both changes to 10.4.

linawolf:
Backporting between master and 10.4 cannot be done automatically easaly due to the overhaul.

linawolf:
Therefore I manualy inserted the changes from master into this PR. Should we merge it like that, @froemken ?

mbrodala (approved):

froemken:
@linawolf Perfect. Can be merged. The explanation is much better than my ;-)

PR #354, opened 2021-02-15
Fix "attention box"

franzholz:
Reverts TYPO3-Documentation/TYPO3CMS-Reference-TCA#353

It still does not work.


mbrodala:
I wouldn't revert and then attempt another fix but actually do the fix now. ;-)

mbrodala (approved):
I'm pretty sure that it will work now. Sorry for not noticing the last time.

mbrodala:
Yup, works now:

image

PR #356, opened 2021-02-15
Array $this->data

franzholz:


mbrodala:
Why this change? The current text is actually just fine ...

franzholz:
You have a big black double quote symbol. This is only for quotes.

The array $this->data provides the following data:

        The row of the currently edited record in $this->data['databaseRow']
        The configuration from the TCA in $this->data['parameterArray']['fieldConf']['config']
        The name of the input field in $this->data['parameterArray']['itemFormElName']
        The current value of the field in $this->data['parameterArray']['itemFormElValue']
        The JavaScript function to be called on changing the field in $this->data['parameterArray']['fieldChangeFunc']

In order ...

mbrodala:
OK, so reducing the indentation would be enough and we can keep the current text.

franzholz (commented):
missing part of ::

Documentation/ColumnsConfig/Type/User/Index.rst
```diff
@@ -161,8 +161,10 @@ implementing a rendering. See :ref:`FormEngine docs
          <t3core:security-xss>`.
 
 
-   The array :php:`$this->data` provides the following data:
```
mbrodala, line 164: Please restore the original text ...

Documentation/ColumnsConfig/Type/User/Index.rst
```diff
@@ -161,8 +161,10 @@ implementing a rendering. See :ref:`FormEngine docs
...
-
+   Array :php:`$this->data`: ...
+   
+   ::
+   
       * The row of the currently edited record in
```
mbrodala, line 168: ... and just reduce the indentation here.

mbrodala (changes requested):

Documentation/ColumnsConfig/Type/User/Index.rst
franzholz: The original text gives no information. Why not make it shorter?

Documentation/ColumnsConfig/Type/User/Index.rst
franzholz: Shouldn't the block remain?

Documentation/ColumnsConfig/Type/User/Index.rst
mbrodala: The current version is better suited for humans. ;-)

Documentation/ColumnsConfig/Type/User/Index.rst
mbrodala: Actually no, a plain list is enough. (Technically a table would be better but that's too much change at once.)

Documentation/ColumnsConfig/Type/User/Index.rst
```diff
@@ -161,18 +161,18 @@ implementing a rendering. See :ref:`FormEngine docs
...
-        :php:`$this->data['databaseRow']`
-      * The configuration from the TCA in
-        :php:`$this->data['parameterArray']['fieldConf']['config']`
-      * The name of the input field in
-        :php:`$this->data['parameterArray']['itemFormElName']`
-      * The current value of the field in
-        :php:`$this->data['parameterArray']['itemFormElValue']`
-      * The JavaScript function to be called on changing the field in
-        :php:`$this->data['parameterArray']['fieldChangeFunc']`
+   The array :php:`$this->data` consists the following parts:
```
mbrodala, line 164:
```suggestion
   The array :php:`$this->data` consists of the following parts:
```

mbrodala (approved):

PR #366, opened 2021-03-04
Fix indentation and line length

### LINKED ISSUE #365: blockquote on DateTime input => eval
The indents should not be there

https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Input/DateTime/Properties/Eval.html

marble:
Resolves #365


marble:
@linawolf FYI: I checked that in without review, wanted to see if the 'Resolves #365' works. And, yes: It did work.

PR #368, opened 2021-03-04
Fix toctree nesting

### LINKED ISSUE #364: "How to use the styleguide extension" shows up twice in the menu
I don't know exactly what is the problem there, but it looks strange:

image

https://docs.typo3.org/m/typo3/reference-tca/master/en-us/Examples/StyleGuide.html#styleguide

marble:
Resolves #364

In pages with a toctree directive the document title should be the only headline.

### DIFF
--- a/Documentation/Examples/About.rst
+++ b/Documentation/Examples/About.rst
@@ -1,0 +1,67 @@
+.. include:: /Includes.rst.txt
+
+.. _tca_examples:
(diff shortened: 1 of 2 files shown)


marble:
@linawolf Damn, was too quick, maybe I shouldn't have merged this. Don't know if that conflicts with your automations

PR #372, opened 2021-03-11
from a record of one table

franzholz:


Documentation/ColumnsConfig/Type/Group/Index.rst
```diff
@@ -6,23 +6,23 @@
 Group fields
 ============
 
-The group element in TYPO3 makes it possible to create references to records from multiple tables in the system.
+The group element (:php:`type' == 'group'`) in TYPO3 makes it possible to create references from a record of one table to many records from multiple tables in the system.
```
mbrodala, line 9:
```suggestion
The group element (:php:`type' => 'group'`) in TYPO3 makes it possible to create references from a record of one table to many records from multiple other tables in the system.
```

mbrodala (changes requested):

Documentation/ColumnsConfig/Type/Group/Index.rst
franzholz: It can also be a related between to tt_content tables. So both tables can be the same.

franzholz (commented):
Help the programmer to find the enable method.

Documentation/ColumnsConfig/Type/Group/Index.rst
mbrodala: True, I've adjusted the suggestion to make this clearer.

franzholz: instead of "The foreign table(s)"
better:
The foreign table

This would be understandable.

mbrodala: Well, there can be many foreign tables ...

franzholz: But please remove the (s). This does not look good.

mbrodala: Sure, done now. :+1:

mbrodala (approved):

PR #376, opened 2021-03-26
[Feature] #57082 - New TCA type “language”

linawolf:
refs TYPO3-Documentation/TYPO3CMS-Reference-CoreApi#1353


Documentation/ColumnsConfig/Type/Language/Introduction.rst
```diff
@@ -0,0 +1,32 @@
+.. include:: /Includes.rst.txt
+.. _columns-languge-introduction:
+
+============
+Introduction
+============
+
+A new TCA field type called `language` has been added to TYPO3 Core. It's main
```
susannemoog, line 8: Please adjust the text from the changelog to avoid references to "new" - in case of this introduction I would split it in two parts: one that is a real introduction (what is this field, what does it do) and one that describes the history for people who are working with TYPO3 since a while longer. This allows users that are new or are "just here for the info" to skim the chapter and find what they need quickly.

Documentation/ColumnsConfig/Type/Language/Properties/Index.rst
```diff
@@ -0,0 +1,12 @@
+.. include:: /Includes.rst.txt
+
+=========================
+Language field properties
+=========================
+
+Since the new TCA type is mostly based on the `type=select` internally, most
```
susannemoog, line 7: Here too, please rewrite these and avoid "new"

susannemoog (approved):

PR #385, opened 2021-04-13
[TASK] Ask information about change in location Extending TCA (7.6)

sypets:
"Extending TCA" was moved to "TYPO3 Explained" since version 8.7.
The outdated pages in TCA reference are still often listed as top results
for search.

This is a fix for directing users to the newer information.

Depends on: TYPO3-Documentation/TYPO3CMS-Reference-CoreApi#1371

Releases: 7.6

### DIFF
--- a/Documentation/ExtendingTca/Examples/Index.rst
+++ b/Documentation/ExtendingTca/Examples/Index.rst
@@ -6,6 +6,14 @@
 Customization examples
 ^^^^^^^^^^^^^^^^^^^^^^
 
+.. note::
+
+   This page is for an outdated TYPO3 version (7.6). For newer versions, you
+   can find the information in:
+
+   *  TYPO3 Explained: :ref:`t3coreapi:extending-tca` (7.6)
+
+
 Many extracts can be found throughout the manual, but this section
 provides more complete examples.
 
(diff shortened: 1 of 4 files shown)


sypets:
This is in combination with https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/1371

I searched for "typo3 tca add column", then the 7.6 page was in the top results:https://docs.typo3.org/m/typo3/reference-tca/7.6/en-us/ExtendingTca/Examples/Index.html


I hope it might help a little to guide people towards the information they are looking for. 

I don't plan to add a lot of these - because it does not really feel like a great solution, more like a hack and it would probably be better to solve these problems in another way.

It also might be helpful to figure out why in some case the old pages are (still) ranked higher (after all this time).

sypets:
Apart from that - I think moving the "Extending TCA" to TYPO3 Explained was not a good idea. Apart from the general problems of restructruging with broken links and not being able to switch with the version selector.

I think it would be better to have it all in one manual - the introduction, examples and the reference - as is the case also in TypoScript reference (at least a chapter for using and setting is there) and in TSconfig.

linawolf (approved):

PR #386, opened 2021-04-16
fix bullet list

franzholz:


linawolf:
The problem of these bullet list lies not in the symbol (* and - are both allowed) but in the spaces. See:
https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/Lists.html

linawolf (approved):

PR #411, opened 2021-06-24
[BUGFIX] Fix title of input type=none format property

dogawaf:


susannemoog (approved):

susannemoog:
10.4 only, as this is moved (and fixed) in master.

PR #412, opened 2021-06-30
[TASK] Remove all showRemovedLocalizationRecords occurrences

nhovratov:
This TCA option is not used by the core since v7 anymore.


lolli42:
Confirmed. this is fine.
See as core cleanup: https://review.typo3.org/c/Packages/TYPO3.CMS/+/69665

PR #420, opened 2021-07-16
[BUGFIX] Fix inverted description

nhovratov:
The TCA option localizeReferencesAtParentLocalization only works if
the field is NOT stored in a MM table.

Also add the select field as a possible type, as it works with all
fields defining foreign_table.


waldhacker1:
@nhovratov is right, see https://github.com/TYPO3/TYPO3.CMS/blob/637fb942211dbf4c277c1c720e6757acee7cae54/typo3/sysext/core/Classes/DataHandling/DataHandler.php#L3690

susannemoog (approved):

PR #432, opened 2021-08-17
[BUGFIX] Remove double "read"

simonschaufi:


Documentation/ColumnsConfig/Type/None/Index.rst
```diff
@@ -17,7 +17,7 @@ Characteristics of `none`:
 
 * The DataHandler discards values send for type none and never persists or updates them in the database.
 * Type none is the only type that does **not** necessarily need a database field.
-* Type none fields does have a default renderType in FormEngine that displays the value as read read only
+* Type none fields does have a default renderType in FormEngine that displays the value as read only
```
mbrodala, line 20:
```suggestion
* Type none fields does have a default renderType in FormEngine that displays the value as readonly
```

mbrodala (changes requested):

Documentation/ColumnsConfig/Type/None/Index.rst
simonschaufi: @mbrodala There is another occurrence of it in "If no database field exists for none fields, an empty read only input field is rendered by default." Which then should also be adjusted, or?

mbrodala: Yes, please adjust all occurences.

linawolf (approved):

linawolf:
@simonschaufi If you need your changes backportet just write Releases: master, 10.4, 9.5 no need to make a separate PR for each version

PR #450, opened 2021-09-25
[Deprecation] Deprecate inline JavaScript in fieldChangeFunc

linawolf:


linawolf:
@lolli42 should we just remove this part from the example or create a new, more elaborate example based on OnFieldChangeInterface?

PR #466, opened 2021-10-31
[TASK] Clarify checkbox default value

nhovratov:
- Explain the binary representation with help of a table
- Explain how to find out the desired decimal value.
- Add missing default value in TCA example.


linawolf:
The code snippets get generated automatically from the extension styleguide. Therefore changing them will be overriden lateron.

linawolf (approved):

nhovratov:
Thanks for the merge! I added the `default` part manually to the code snippet. Should we then exclude it from future updates?

PR #470, opened 2021-11-10
[FEATURE] Add example for symmetric field in inline TCA

linawolf:


Documentation/ColumnsConfig/Type/Inline/Properties/SymmetricField.rst
```diff
@@ -14,3 +14,10 @@ symmetric\_field
    This works like :ref:`foreign_field <columns-inline-properties-foreign-field>`, but in case of using
    bidirectional symmetric relations. :code:`symmetric_field` defines in which field on
    the :ref:`foreign_table <columns-inline-properties-foreign-table>` the uid of the "other" parent is stored.
+
+Example
+=======
+
+.. include:: /CodeSnippets/InlineMnSymetricInline1.rst.txt
```
susannemoog, line 21: Symmetric is spelled with two m (see text block before this) - we should probably change that at wherever the source of these is.

linawolf:
continued in https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/490

PR #478, opened 2021-11-19
Minor typo in LanguageField.rst breaks link

lolli42:

PR #480, opened 2021-11-20
Minor typo in LanguageField.rst breaks link (#478)

lolli42:


susannemoog:
The references themselves all contain the same typo (`languge`), but I'll leave that in, for now, to avoid breaking links and making this bigger than it needs t be.

susannemoog (approved):

PR #487, opened 2021-11-23
changes

franzholz:
Question:
Is the Overrides folder required for "category"? 
Which field is added by TYPO3 to the database?

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.5/Breaking-78384-FrontendIgnoresTCAInExtTables.html#migration

### DIFF
--- a/Documentation/ColumnsConfig/Type/Category/Index.rst
+++ b/Documentation/ColumnsConfig/Type/Category/Index.rst
@@ -16,7 +16,7 @@ Category
 While using the type :php:`category`, TYPO3 takes care of generating the
 necessary TCA configuration.
 Developers only have to define the TCA column and add :php:`category` as the
-desired TCA type.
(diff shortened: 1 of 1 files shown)


DavidBruchmann:
no if you have an own table-definition you can define it directly there.
The linked example overrides tt_content and needs therefore be placed in the Overrides folder.
The corresponding field depends on the definition, it can be an existing field "categories" but could also be a new one and independent of any existing fields (This logic might be hard for usability and the developer, but in general everything is open).

franzholz:
Does this mean that if there is not field "categories" in "example_table", then TYPO3 will read the TCA in order to add this field and insert it into the "example_table"?

DavidBruchmann:
no, it had to be added in the database too and that's out of scope of TYPO3

DavidBruchmann:
beside that you're not bound to the name "categories" you could name it "tomatoes" or anything else too

linawolf (approved):

PR #506, opened 2021-12-29
[BUGFIX] Fix blockquotes in datetime properties

nhovratov:
Releases: main, 11.5


linawolf (changes requested):
the directives like
`:type: string` make no sense without the main directive. I think, the confval directive is missing here and should be added instead of removing the indentations.

linawolf (approved):

linawolf:
fixed it

PR #517, opened 2022-01-07
[BUGFIX] Remove wrongly placed deprecation notice

nhovratov:
The deprecation for foreign_table => sys_language does
only apply to type=select.

Releases: main, 11.5


linawolf:
Usage of the table  sys_language has been deprecated overall and will be removed. What is wrong about that?

nhovratov:
Type inline is usually not used for sys_language relations. And the migration to type=language is a direct replacement for type=select with special=language. So it makes no sense to mention it for type=inline.

The removal of the sys_language table is a general deprecation and not strictly related to the hint.

linawolf (approved):

Issue #404, opened 2021-06-13
How to show the array path in the TCA ref?


sypets:
## Problem

I often have the problem, I land on one page and then I don't know where the option is in the `$GLOBALS['TCA']`  arrray.

This is especially true, because of a change made from 10.4 to master where the menu item config is no longer called `['columns'][*]['config']` but `Field types (config > type)` and more descriptive menu items were used in general instead of the technical representation.

* 10.4: https://docs.typo3.org/m/typo3/reference-tca/10.4/en-us/ColumnsConfig/Index.html#
* master: https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Index.html

## Examples:

* config-> type is actuallly $GLOBALS['TCA'][$table]['columns']['config']['type'] - as this is on the same menu level as the other top level options like 'columns', 'ctrl' and 'interface', this can be confusing. (I found it better, the way it was before, e.g. see 10.4 version where `['columns'][*]['config']` was visible in the menu

## Possible solution

1a. show path or partial path in menu (can be a bit long)
1b OR at least show `['columns'][*]['config']` as it was before
2. on top of each page, show the full path, e.g. [$table]['columns']['config']['type']['authMode'] on page authMode: https://docs.typo3.org/m/typo3/reference-tca/10.4/en-us/ColumnsConfig/Type/selectSingle.html#authmode

PR #520, opened 2022-01-22
[DOCS] Add TCA path information

linawolf:
Add information about where exactly the configuration is located in the TCA array

releases: main, 11.5

Fixes: #404

### DIFF
--- a/Documentation/ColumnsConfig/CommonProperties/FieldControl.rst
+++ b/Documentation/ColumnsConfig/CommonProperties/FieldControl.rst
@@ -5,6 +5,7 @@
 fieldControl
 ============
 
+   :Path: $GLOBALS['TCA'][$table]['columns'][$field]['config']
 .. confval:: fieldControl
 
    :type: array
--- a/Documentation/ColumnsConfig/CommonProperties/FieldControl.rst
+++ b/Documentation/ColumnsConfig/CommonProperties/FieldControl.rst
@@ -5,9 +5,9 @@
 fieldControl
 ============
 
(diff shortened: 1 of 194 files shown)


Documentation/ColumnsConfig/CommonProperties/FieldControl.rst
```diff
@@ -5,6 +5,7 @@
 fieldControl
 ============
 
+   :Path: $GLOBALS['TCA'][$table]['columns'][$field]['config']
```
nhovratov, line 8: Should be inside `confval`

Documentation/ColumnsConfig/Type/Group/Properties/FieldControl.rst
```diff
@@ -13,6 +13,7 @@ The field of type group can enable all common :ref:`field control options
 
 .. confval:: elementBrowser
 
+   :Path: $GLOBALS['TCA'][$table]['columns'][$field]['config']
```
nhovratov, line 16: Add it for insertClipboard too?

nhovratov (changes requested):
Looks good, this gives a nice breadcrumb like orientation.
Found only one path outside the confal directive and another one, which can be added.
Besideds that +1 from me.

nhovratov:
Also, a "Fixes" can be added to the commit message for #404:

`Fixes: #404`

marble (approved):
I totally apreciate having that Path information. I'm not a hundred percent sure though the name will just stay "Path" if some built in functionatility is to be linked to that option. But this can surely be easily adapted to by some editorial measures.

PR #523, opened 2022-01-30
Override not overwrite configuration values

linawolf:
as the original still exists.

See https://english.stackexchange.com/questions/88400/when-to-use-override-and-overwrite

releases: main, 11.5


flicstar (approved):

flicstar:
Looks good to me +1 but I didn't check the "correctness" of the statements. I did follow the conversation about this in Slack though, so I assume you know what you're doing 😉

PR #535, opened 2022-02-22
[BUGFIX] Fix header level

nhovratov:
Releases: main, 11.5


linawolf (changes requested):
We need to keep the escape of the underscore here as that is a special sign in rst

nhovratov:
> We need to keep the escape of the underscore here as that is a special sign in rst

Where exactly does this become a problem? In my local rendering everything seems fine.

linawolf (approved):

PR #536, opened 2022-02-26
Do not use "record type" as the title.

franzholz:
I want to search for [$GLOBALS['TCA']['tx_news_domain_model_news']['ctrl']['type'].
It is only searchable if the title matches exactly the TCA array key name!

### DIFF
--- a/Documentation/Ctrl/Properties/Type.rst
+++ b/Documentation/Ctrl/Properties/Type.rst
@@ -1,9 +1,9 @@
 .. include:: /Includes.rst.txt
 .. _ctrl-reference-type:
 
-============
-Record types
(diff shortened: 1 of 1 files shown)


linawolf:
I don't see your point, you can search for "type" just fine:

image

linawolf (changes requested):
No change is needed to make this searchable

franzholz:
When you have a TCA array in your PHP code, then you will open the Documentation page and search in the left menu for the entries by the key name of the TCA array. All other entries have their own key entry. Only this one has another title. This makes it complicate.

linawolf (approved):

PR #545, opened 2022-03-18
Raise attention on ctrl/sortby

mbrodala:
The current markup/rendering does not really convey the importance of using this field propertly:

image

PR #546, opened 2022-03-18
Raise attention on inline foreign_sortby

mbrodala:
This mirrors ctrl/sortby since the same mistakes can be done here.

See https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/545


linawolf:
releases: main. 11.5

mbrodala:
Thanks. This hopefully raises enough attention on these (admittedly oddly named) options.

PR #564, opened 2022-04-04
[Feature] #97013 - New TCA type “email”

linawolf:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-97013-NewTCATypeEmail.html

refs https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/1624


nhovratov:
`eval => 'email'` should be removed from type `input`. Maybe with an added migration hint? There will be more of those: datetime and number.

PR #569, opened 2022-04-05
[TASK] Add old inputLink reference to new link type

nhovratov:


Documentation/ColumnsConfig/Type/Link/Index.rst
```diff
@@ -1,5 +1,6 @@
 .. include:: /Includes.rst.txt
 
+.. _columns-input-renderType-inputLink:
```
linawolf, line 3: should we not rather fix all areas that still have the old link, as they might very well describe the old renderType?

nhovratov: I don't know. The type link is a direct replacement for renderType inputLink. We would just break some links, if we omit it.

linawolf: Ok, lets do it then

linawolf (approved):

PR #571, opened 2022-04-05
[TASK] Cleanup input type

nhovratov:
The TCA type "input" no longer has any renderType.
Because of this, a lot can be simplified.
We do no longer have to refer to render types.
Examples are moved to the entry page.
Leftover options are orphaned and linked to the new types.


Documentation/ColumnsConfig/Type/Input/Properties/LinkPopup.rst
```diff
@@ -7,29 +7,7 @@
 linkPopup
```
nhovratov, line 7: This can be removed and redirected to the equivalent `columns-link-properties-appearance`

linawolf (approved):

Issue #599, opened 2022-06-01
Replace class name in string with FQCN with ::class


sypets:
Moved here from #252

----

Class name in ext_localconf.php : replace string with path::class 

not:
`$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['tce']['formevals']['Vendor\\Extension\\Evaluation\\ExampleEvaluation'] = '';`

but:
    `$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['tce']['formevals'][\Vendor\Extension\Evaluation\ExampleEvaluation::class] = '';`

https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/Input/Properties/Eval.html

Issue #252, opened 2019-12-05
eval / custom class example with class path as string does not work with double-slashes for FlexForm


sypets:
I am using the example for ["eval" => "Vendor ..."](https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/Input/Properties/Eval.html), but using this in FlexForms

The string as specified in the example with double-slashes does not work in my example. String with single-slashes does, e.g.

not:
`'eval' => 'trim,Vendor\\Extension\\Evaluation\\ExampleEvaluation,required'`

but:
`'eval' => 'trim,Vendor\Extension\Evaluation\ExampleEvaluation,required'`

(However, as stated above this was tested with FlexForm, not with TCA directly)

remonator:
Specifying the class name in `ext_localconf.php` as you did is also necessary when using without FlexForms (using 10.4.9).

linawolf:
Can be found at https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Input/Properties/Eval.html now, Move Code to styleguide extension and make sure it is running.

PR #600, opened 2022-06-01
Use ::class with class name instead of static string

sypets:
Resolves: #599
Resolves: #252
Releases: master, 11.5, 10.4

----

Not in commit: 

For v12, the fully qualified class name in ext_localconf.php can also be changed to the short class name (`ExampleEvaluation::class`) and "use" statements added. That can be done in separate PR/commit, so this patch here can be backported more easily.

### DIFF
--- a/Documentation/ColumnsConfig/Type/Input/Properties/Eval.rst
+++ b/Documentation/ColumnsConfig/Type/Input/Properties/Eval.rst
@@ -180,7 +180,7 @@ saving the record:
 .. code-block:: php
 
    // Register the class to be available in 'eval' of TCA
-   $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['tce']['formevals']['Vendor\\Extension\\Evaluation\\ExampleEvaluation'] = '';
+   $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['tce']['formevals'][\Vendor\Extension\Evaluation\ExampleEvaluation::class] = '';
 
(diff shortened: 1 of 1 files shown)


Documentation/ColumnsConfig/Type/Input/Properties/Eval.rst
```diff
@@ -190,7 +190,7 @@ saving the record:
       'example_field' => [
          'config' => [
             'type' => 'text',
-            'eval' => 'trim,Vendor\\Extension\\Evaluation\\ExampleEvaluation,required'
+            'eval' => 'trim,required,' . \Vendor\Extension\Evaluation\ExampleEvaluation::class
```
linawolf, line 193: required has been moved to a separate property for V12, so we would have to take your changes as here to version 11 then move required to its own property for version 12

linawolf (changes requested):

Documentation/ColumnsConfig/Type/Input/Properties/Eval.rst
sypets: Ok, what should I do? Maybe it's easier if I create separate PR for each branch or do you want to do this?

linawolf: I think in this case it is better to make one PR for each version. also we cannot backport to 10.4 as the structure changed. Could you do that?

sypets: ok

sypets:
close here - will create separate patches for each version.

PR #601, opened 2022-06-04
Replace class name in static string in eval (v12 / main branch)

sypets:
Use ::class with class name instead of static string in the description
of the property eval.

Additionally, "required" is removed from the string, since this has been
moved to a separate property since v12.

Some captions reflecting the file names are added to the code blocks.
    
For v12, we can also add a "use" to ext_localconf.php to define the
full class name and use the short class name below.

Resolves: #599
Resolves: #252

---

Do not backport this PR - patches for other branch 11.5 will be supplied with a separate PR.

### DIFF
--- a/Documentation/ColumnsConfig/Type/Input/Properties/Eval.rst
+++ b/Documentation/ColumnsConfig/Type/Input/Properties/Eval.rst
@@ -175,22 +175,24 @@ saving the record:
       }
    }
 
-:file:`EXT:extension/ext_localconf.php`:
-
 .. code-block:: php
+   :caption: EXT:extension/ext_localconf.php
+
+   use Vendor\Extension\Evaluation\ExampleEvaluation;
+
+   // ....
 
    // Register the class to be available in 'eval' of TCA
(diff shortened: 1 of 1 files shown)


Documentation/ColumnsConfig/Type/Input/Properties/Eval.rst
```diff
@@ -175,22 +175,24 @@ saving the record:
...
+   // Register the class to be available in 'eval' of TCA
+   $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['tce']['formevals'][ExampleEvaluation::class] = '';
 
 .. code-block:: php
+   :caption: EXT:extension/Configuration/TCA/tx_example_record.php
 
    'columns' => [
       'example_field' => [
          'config' => [
             'type' => 'text',
```
linawolf, line 194: Shall we keep the required option, just to make ppl used to the change?

```suggestion
            'type' => 'text',
            'required' => true,
```

sypets: Is strictly speaking not part of "eval" but I think that is a good idea.

linawolf: The type is not part of eval either, these are just surrounding information for easier grasp

linawolf (approved):

PR #612, opened 2022-06-17
[TASK] Streamline spelling of "color picker"

brotkrueml:


linawolf:
Did you check if there is further misspelling in this reference? We should change them all at once if possible.

brotkrueml:
> Did you check if there is further misspelling in this reference? We should change them all at once if possible.

Yep, searched for "color-picker" in the whole project.

linawolf (approved):

PR #617, opened 2022-07-07
Wrong column type in StartimeEndtime example

ghermens:


linawolf:
That is right, `datetime` was only introduced with v. 12

linawolf (approved):

PR #618, opened 2022-07-07
[BUGFIX] wrong column type in StartimeEndtime example

ghermens:


ghermens:
applies only to 11.5 branch ...

PR #620, opened 2022-07-18
Remove escape of underscore in confval title

linawolf:
Otherwise the backslash will be visible in the output


mbrodala:
Removed the backport label since v11 is actually fine:

| main | v11 |
| --- | --- |
| image | image |

mbrodala (approved):

linawolf:
V 11 is fine since it has not-yet been re-rendered, however on rendering on the new container this becomes an issue right-away

PR #640, opened 2022-09-04
Mandatory fields are actually created automatically

linawolf:
Releases: main, 11.5

### DIFF
--- a/Documentation/BestPractises/CommonFields.rst
+++ b/Documentation/BestPractises/CommonFields.rst
@@ -12,14 +12,14 @@ Mandatory fields
 If the table has a TCA definition, TYPO3 will automatically create the following fields:
 
 :sql:`uid`
-    A unique identifier, integer, auto-incrementing. This field is used as table key
-    and to store inter record relationships.
+    An auto-incrementing unique identifier. This field is used as table key
+    and as a reference in relationships between records.
 
 :sql:`pid`
-    The page the record is situated on or 0 if it is attached to no page
+    The `uid` property of the parent page. The record is situated on this page. This value is 0 if the record is not connected to any page.
 
 There is no separate TCA definition of these fields in the TCA configuration. It is
-not possible to use different names for these fields.
+not possible to use other names for these fields.
 
 Fields used by convention
 =========================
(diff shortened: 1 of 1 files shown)


Documentation/BestPractises/CommonFields.rst
```diff
@@ -9,18 +9,17 @@ Common fields
 Mandatory fields
 ================
 
-All data tables with a TCA definition need to have at least the following
-fields defined in the :file:`ext_tables.sql`:
+If there is a TCA definition for a table TYPO3 will automatically create the following
+fields:
```
mabolek, line 13:
```suggestion
If the table has a TCA definition, TYPO3 will automatically create the following fields:
```

Documentation/BestPractises/CommonFields.rst
```diff
@@ -9,18 +9,17 @@ Common fields
...
 
-.. code-block:: sql
-   :caption: EXT:my_extension/ext_tables.sql
+:sql:`uid`
+    A unique identifier, integer, auto-incrementing. This field is used as table key
+    and to store inter record relationships.
```
mabolek, line 17:
```suggestion
    An auto-incrementing unique identifier. This field is used as table key
    and as a reference in relationships between records.
```

Documentation/BestPractises/CommonFields.rst
```diff
@@ -9,18 +9,17 @@ Common fields
...
 
-   CREATE TABLE tx_myextension_domain_model_something (
-      uid          INT(11)                 NOT NULL AUTO_INCREMENT,
-      pid          INT(11) DEFAULT '0'     NOT NULL,
-   )
+:sql:`pid`
+    The page the record is situated on or 0 if it is attached to no page
```
mabolek, line 20:
```suggestion
    The `uid` property of the parent page. The record is situated on this page. This value is 0 if the record is not connected to any page.
```

Documentation/BestPractises/CommonFields.rst
```diff
@@ -9,18 +9,17 @@ Common fields
...
 
-There is no separate definition of these fields in the TCA configuration. It is
+There is no separate TCA definition of these fields in the TCA configuration. It is
 not possible to use different names for these fields.
```
mabolek, line 23:
```suggestion
There is no separate TCA definition of these fields in the TCA configuration. It is
not possible to use other names for these fields.
```

mabolek (changes requested):

PR #660, opened 2022-10-25
Remove inappropriate example

julianhofmann:
The example does not show the property èxclude`. So it has no added value for the user.

I think 'exclude' is self-explanatory enough that it doesn't need an example/image to illustrate its use.

### DIFF
--- a/Documentation/Columns/Properties/Exclude.rst
+++ b/Documentation/Columns/Properties/Exclude.rst
@@ -18,3 +18,23 @@ exclude
 
    See :ref:`Access lists <t3coreapi:access-options-access-lists>` for more
    about permissions.
+
+Example
+=======
+
+Simple input field
+------------------
+
+.. code-block:: php
+   [
+       'columns' => [
+           'input_2' => [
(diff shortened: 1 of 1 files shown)


linawolf (changes requested):
The example should be replaced, not removed.  I don't think we need a screenshot but a code-snippet would be good, as exclude is put outside the config-section which is not what developers are used to

Documentation/Columns/Properties/Exclude.rst
```diff
@@ -25,6 +25,16 @@ Example
 Simple input field
 ------------------
 
-.. include:: /Images/Rst/Input1.rst.txt
-
-.. include:: /CodeSnippets/Input1.rst.txt
+.. code-block:: php
+   [
+       'columns' => [
```
linawolf, line 30:
```suggestion
.. code-block:: php

   [
       'columns' => [
```

code-block must be followed by an empty line or it does not render properly

linawolf (approved):

PR #661, opened 2022-10-25
Add edge case info about category type

runepiper:
While the documentation states that you can override some fields, I feel the need to indicate that there are edge cases where `type=category` is not suitable. What do you think?

Is it possible to add the hacktoberfest-accepted label?

### DIFF
--- a/Documentation/ColumnsConfig/Type/Category/Index.rst
+++ b/Documentation/ColumnsConfig/Type/Category/Index.rst
@@ -1,4 +1,4 @@
-﻿.. include:: /Includes.rst.txt
+.. include:: /Includes.rst.txt
 
 .. _columns-category:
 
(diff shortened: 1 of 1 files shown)


Documentation/ColumnsConfig/Type/Category/Index.rst
```diff
@@ -29,12 +29,13 @@ The following options can be overridden via :ref:`page TSconfig, TCE form
 *  `readOnly`
 *  `treeConfig`
 
-
 .. note::
 
    It is still possible to configure a category tree with `type=select`
    and `renderType=selectTree`. This configuration will still work, but
    it can in most cases be simplified by using the new :php:`category` TCA type.
+   However if you need to modifiy e. g. the `foreign_table_where` field, you
```
linawolf, line 37: The note sounds a little bumpy now. "replicate the category type" is about the same like "configure a category tree with `type=select`
   and `renderType=selectTree`" all you are adding is one example when that would be neccessary. To it could be added as an example in the sentence above

linawolf (changes requested):
Please adjust the note box

Documentation/ColumnsConfig/Type/Category/Index.rst
runepiper: I hope my updated pull request is better now.

linawolf (approved):

PR #681, opened 2023-01-29
[BUGFIX] Resolve rendering Warnings

linawolf:


DavidBruchmann (approved):
I don't understand all the removals but trust you that you know what you do

linawolf:
I replaced the menus with every entry listed with globs -  a menu that automatically includes all files by wildcard, see https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Menus/MenuHierarchy.html#example-glob-not-hidden-titlesonly

And I removed examples of the usage of inline fields for FAL as with v12 the field of type "file" was introduced and there these examples dont work anymore, see https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/File/Index.html#migration

PR #682, opened 2023-01-29
[FEATURE] Add dbType "json" to TCA type "user"

linawolf:
resolves https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/277


Documentation/ColumnsConfig/Type/User/Index.rst
```diff
@@ -120,7 +120,59 @@ This example is also described in TYPO3 Explained,
...
 
-..  _columns-user-properties-type:
+..  _columns-user-properties-dbType:
+
+dbType
+------
+
+..  versionadded:: 12.2
+    The dbType `json` was introduced to allow storage and usage of JSON data
+    without needing to decode the JSON in each user implementation manually.
```
sbuerk, line 130: I would emphasis that the decoded is only done with core form handling
in the backend. And not if the the data is retrieved in any other way.

FE or custom implementation retrieving the records still *must* decode
the value themself.

If `Connection->insert()` or `Connection->update()` are used, and not
specified types handed over, the database scheme is used and native for
json fields data will then encoded. This does not work for `QueryBuilder`.
Same goes for extbase, which not decoding this yet (retrieving a string).

sbuerk (changes requested):

Documentation/ColumnsConfig/Type/User/Index.rst
linawolf: I removed the confuusing half sentence here and added a note box with this information

PR #694, opened 2023-03-22
Correct overrideable options for the category type

FLeudts:
The "foreign_table_where option" is only set if there isn't already a value set
https://review.typo3.org/76415
https://github.com/TYPO3/typo3/blob/a93084546bb22daa7f170b8d628d3041ecd324b7/typo3/sysext/core/Classes/Preparations/TcaPreparation.php#L87

### DIFF
--- a/Documentation/ColumnsConfig/Type/Category/Properties/ForeignTableWhere.rst
+++ b/Documentation/ColumnsConfig/Type/Category/Properties/ForeignTableWhere.rst
@@ -1,17 +1,19 @@
 .. include:: /Includes.rst.txt
-.. _columns-category-properties-foreign-table-where::
+.. _columns-category-properties-foreign-table-where:
 
-=============
+===================
 foreign_table_where
-=============
+===================
 
-.. confval:: foreign_table_where
+.. confval:: foreign_table_where (type => category)
 
    :Path: $GLOBALS['TCA'][$table]['columns'][$field]['config']
    :type: string (SQL WHERE)
    :Scope: Proc. / Display
+   :Default:  :sql:`AND {#sys_category}.{#sys_language_uid} IN (-1, 0)`
 
    The items from :ref:`foreign_table <columns-select-properties-foreign-table>`
-   are selected with this WHERE-clause.
+   are selected with this :sql:`WHERE` clause. If set make sure to append the default value
+   to your query.
    See also :ref:`property foreign_table_where of select field
    <columns-select-properties-foreign-table-where>`.


Documentation/ColumnsConfig/Type/Category/Index.rst
```diff
@@ -24,6 +24,7 @@ The following options can be overridden via :ref:`page TSconfig, TCE form
 <t3tsconfig:pageTsConfigTceFormConfig>`:
 
 *  `size`
+*  `foreign_table_where`
```
brotkrueml, line 27: This seems not to be correct: https://github.com/TYPO3/typo3/blob/main/typo3/sysext/backend/Classes/Form/Utility/FormEngineUtility.php#L59

FLeudts: What you are referencing is the ability to overwrite TCA in the TSconfig file. What this documentation is talking about is overwriting TCA presets in the TCA definition.

FLeudts: Sorry, I realize what I said is wrong. It is of course talking about the TSconfig file in the first part, but I was confused because of the sentence in the note section. The fact that the list is inserted between two parts that talk about the TCA file never made me question that introduction sentence. I always skipped reading it apparently. Sorry for the confusion.

Documentation/ColumnsConfig/Type/Category/Properties/ForeignTableWhere.rst
```diff
@@ -0,0 +1,17 @@
+.. include:: /Includes.rst.txt
+.. _columns-category-properties-foreign-table-where::
+
+=============
```
brotkrueml, line 4: The "=" above and below must be at least as long as the header text, but it can be longer. Otherwise there will be a rendering error.

Documentation/ColumnsConfig/Type/Category/Properties/ForeignTableWhere.rst
```diff
@@ -0,0 +1,17 @@
...
+
+.. confval:: foreign_table_where
+
+   :Path: $GLOBALS['TCA'][$table]['columns'][$field]['config']
+   :type: string (SQL WHERE)
+   :Scope: Proc. / Display
+
+   The items from :ref:`foreign_table <columns-select-properties-foreign-table>`
+   are selected with this WHERE-clause.
```
brotkrueml, line 15:
```suggestion
   are selected with this :sql:`WHERE` clause.
```

brotkrueml (changes requested):
I haven't used category by myself, so can't say if this is okay.

But from reading the code, I see the following:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Preparations/TcaPreparation.php#L87

So, when using `foreign_table_where` you override the default one with the sys_language constraint. Is this correct? Then the default value should be mentioned, also when using the `foreign_table_where` yourself you should add the default value to your constraint to avoid possible problems.

Or do I overlook something?

Documentation/ColumnsConfig/Type/Category/Properties/ForeignTableWhere.rst
FLeudts: I added your requested changes, added the default value and added a sentence in order to make it clear that the default value should be kept. 
This time I checked and the documentation rendered without warnings for me.

brotkrueml (approved):

brotkrueml:
Looks good for me. As I am not into categories, someone else may have a look before merging.

@FLeudts Thanks for your controbution

linawolf:
@FLeudts thanks for taking care of this issue

PR #707, opened 2023-06-01
correct extensionKey for system extension with example

DavidBruchmann:
"EXT:reaction" was plain wrong, as it's an outdated extension. So the module name was mistaken as extension name. The new system extension is "webhooks" and consists of 2 modules: reactions and webhooks.

### DIFF
--- a/Documentation/ColumnsConfig/Type/User/Index.rst
+++ b/Documentation/ColumnsConfig/Type/User/Index.rst
@@ -155,11 +155,11 @@ dbType
 dbType = json example
 ~~~~~~~~~~~~~~~~~~~~~
 
-The system extension :t3ext:`webhooks` uses a TCA field of type `user` with
+The system extension :t3ext:`reactions` uses a TCA field of type `user` with
 the dbType `json`:
 
 ..  code-block:: php
-    :caption: :t3src:`webhooks/Configuration/TCA/Overrides/sys_reaction_create_record.php` (excerpt)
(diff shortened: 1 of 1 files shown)


Documentation/ColumnsConfig/Type/User/Index.rst
```diff
@@ -155,7 +155,7 @@ dbType
 dbType = json example
 ~~~~~~~~~~~~~~~~~~~~~
 
-The system extension :t3ext:`reaction` uses a TCA field of type `user` with
+The system extension :t3ext:`webhooks` uses a TCA field of type `user` with
```
brotkrueml, line 158: Below is a code example which references EXT:reaction. Also the table sys_raction is referenced. Please check and correct the according places, if applicable.

brotkrueml (changes requested):

Documentation/ColumnsConfig/Type/User/Index.rst
DavidBruchmann: I think the reference to table `sys_reaction` is correct.

brotkrueml:
I am not sure if this is correct:

    LLL:EXT:webhooks/Resources/Private/Language/locallang_db.xlf:sys_reaction.fields

Also the source of the field. This is a real example from reactions extensions, I assume. So, using "webhooks" may be wrong.

DavidBruchmann:
here you go: https://extensions.typo3.org/extension/reaction

DavidBruchmann:
thanks @brotkrueml , your reluctance avoided a mess

PR #708, opened 2023-06-03
Correct extension key for system extension with example (#707)

brotkrueml:


brotkrueml:
Forward-port from #707

linawolf (approved):

PR #709, opened 2023-06-03
Add important note about custom formengine elements

derhansen:
Added a note, that custom formengine elements must contain valid HTML, since unexpected problems may occur if not.

Details: https://forge.typo3.org/issues/100943


brotkrueml (approved):

brotkrueml:
@derhansen Thanks for taking care and your contribution.

PR #712, opened 2023-06-06
[BUGFIX] Correct examples for $GLOBALS['TCA'][$table]['ctrl']['security']

brotkrueml:
The Configuration/TCA/table.php files return an array with the TCA definition. Therefore, the examples should reflect this.

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/Ctrl/Properties/Security.rst
+++ b/Documentation/Ctrl/Properties/Security.rst
@@ -16,13 +16,15 @@ security
     :Scope: Display
 
 
-    Array of sub-properties. This is used in the Core for the :sql:`sys_file` table:
+    Array of sub-properties. This is used, for example, in the Core for the
+    :sql:`sys_file` table:
 
     ..  code-block:: php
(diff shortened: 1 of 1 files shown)


Documentation/Ctrl/Properties/Security.rst
```diff
@@ -21,8 +21,9 @@ security
     ..  code-block:: php
         :caption: EXT:core/Configuration/TCA/sys_file.php
```
linawolf, line 22: I would suggest to make an example on how to use this in my_extension in an override instead. This also removes the nessesarity of comenting out parts etc

brotkrueml: I will add an override example as addition. In custom extensions you have mostly no overrides of your own TCA, but instead integrate it into the default TCA of the table.

linawolf: Then I would still make an example with an own extension

linawolf (approved):

PR #726, opened 2023-06-22
Fix double "only" in AdminOnly.rst

nhovratov:


brotkrueml (approved):

brotkrueml:
@nhovratov Good catch. Thanks for your contribution.

PR #734, opened 2023-06-23
Point at where to learn more about the editor types

### LINKED ISSUE #387: RTE chapter is missing some key points for beginners
A colleague pointed me to https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Text/Index.html and hat troubles configurating an RTE for a text field.

Problems he had:
- The difference of the t3editor and the RTE
- To get an rte ` 'enableRichtext' => true` is needed but this is not documented here

linawolf:
resolves #387
releases: main, 12.4, 11.5


FlorianT03 (approved):
I don't see mistakes. Thanks! :)

PR #750, opened 2023-08-03
[BUGFIX] Add value tag for checkbox FlexForm example

sbusemann:
Without the value tag the data are not persisted 
See https://forge.typo3.org/issues/101476#change-498187


abeutel (commented):
@sbusemann Did not test it yet but have a question: does it make sense to have four checkboxes in this example with empty (= no) value (would this work in some way?)

sbusemann:
If you don't provide an empty value tag, the given data for a "checked" checkbox is not stored - see  https://forge.typo3.org/issues/101476#change-498187

linawolf (approved):

PR #752, opened 2023-08-10
Fixed typo in ValuePicker example

sk-typoconsult:
Fixed typo in ValuePicker example (https://docs.typo3.org/m/typo3/reference-tca/12.4/en-us/ColumnsConfig/Type/Number/Properties/ValuePicker.html)

Forgot a "," at the end of the 'mode' => '' line.

### DIFF
--- a/Documentation/ColumnsConfig/Type/Number/Properties/ValuePicker.rst
+++ b/Documentation/ColumnsConfig/Type/Number/Properties/ValuePicker.rst
@@ -35,7 +35,7 @@ Example
       'label' => 'Number field',
       'config' => [
          'type' => 'number',
-         'mode' => ''
+         'mode' => '',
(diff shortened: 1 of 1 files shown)


brotkrueml (approved):

linawolf:
Thanks a lot for catching this! I just checked and the problem is still present in the main branch. I will create a follow up applying your fix to main. 

For the next time you can save us some work by making your PRs directly on main. We do have automatic backport services that we can activate then. Just tell us something like 

releases: main, 12.4

brotkrueml:
@sk-typoconsult Good catch! Thanks for your contribution :+1:

linawolf (approved):

sk-typoconsult:
Cheers! Will remember that, and thanks!

PR #769, opened 2023-09-08
Ctrl has two main categories, but mentioned three - Update Index.rst

Alagts:
It is confusing for the reader to see three points after mentioning that there are two main categories


linawolf (approved):
Thanks for catching this!

linawolf:
@Alagts you made your changes on branch 12.4 here. That is why  the automatic backport failed. Please, if a change applies to all versions, make your changes on the main branch.

PR #776, opened 2023-09-09
[TASK] Streamline rst files

brotkrueml:
Releases: main, 12.4


brotkrueml:
@linawolf Do you have any clue, why the rst validating is failing:

    no include |              |                   | Columns/Index.rst 
    insert '..  include:: /Includes.rst.txt' in first line of the file

The include is there. I adapted the regex from `\\.\\.` to `\.\.` so it works in regex101:
https://regex101.com/r/sUlkP3/1

But either way, some files fail.

linawolf:
I think I would  not put work into this before the switch to the new parser, which will also allow us validations

brotkrueml:
@linawolf Okay, I removed the workflow part and just streamline now the rst changes. Please have a look again.

linawolf (approved):

PR #783, opened 2023-09-19
[TASK] Fix wording in language history

brotkrueml:
Releases: 12.4, 11.5


brotkrueml:
This page is scheduled to be removed in main (v13) in a follow-up PR, so change is only necessary in v12.4 and v11.5 branches.

brotkrueml:
Nope, we are leaving it in v13 also, as the migration is still in place.

PR #788, opened 2023-09-26
Use proper "quoted" term for PAGE_TSCONFIG_STR

mbrodala:
See https://github.com/TYPO3/typo3/blob/bf477e60fe53d0d7641834f2fed6fca1a6766087/typo3/sysext/backend/Classes/Form/FormDataProvider/AbstractItemProvider.php#L769


sarahmccarthy123 (approved):
Looks fine

PR #807, opened 2023-10-09
Fix wrong reference to language file

derhansen:


brotkrueml:
@derhansen Good catch! Thanks for your contribution :+1:

PR #829, opened 2023-10-22
[FEATURE] #101843 - Allow configuration of color palettes

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/650
Releases: main

### DIFF
--- a/Documentation/ColumnsConfig/Type/Color/Index.rst
+++ b/Documentation/ColumnsConfig/Type/Color/Index.rst
@@ -22,8 +22,7 @@ The TCA type :php:`color` should be used to render a JavaScript-based color pick
 ..  versionadded:: 13.0
 
 :ref:`Color palettes <t3tsconfig:pagecolorpalettes>` can be defined via
-:ref:`page TSconfig <t3tsconfig:setting-page-tsconfig>`. This way, for example,
-colors defined in a corporate design can be made accessible in an easy way.
+:ref:`page TSconfig <t3tsconfig:setting-page-tsconfig>`. With this method, the user can select predefined corporate design colors with a colorpicker instead of typing in RGB values.
 
 Examples
 ========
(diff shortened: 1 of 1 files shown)


Documentation/ColumnsConfig/Type/Color/Index.rst
```diff
@@ -19,6 +19,12 @@ Color
 
 The TCA type :php:`color` should be used to render a JavaScript-based color picker.
 
+..  versionadded:: 13.0
+
+:ref:`Color palettes <t3tsconfig:pagecolorpalettes>` can be defined via
+:ref:`page TSconfig <t3tsconfig:setting-page-tsconfig>`. This way, for example,
+colors defined in a corporate design can be made accessible in an easy way.
```
linawolf, line 26: I don't really get the last sentence?

brotkrueml: The user can just click on the color instead of enter a RGB value manually.

linawolf: colors [...] can be made accessible could also be interpreted as "barrierefrei gemacht werden" this is what caused by confugsion. @sarahmccarthy123 you have an idea of how to rephrase this?

sarahmccarthy123: Maybe "can be selected by a simple click" rather than "can be made accessible in an easy way"

linawolf, line 26:
```suggestion
:ref:`page TSconfig <t3tsconfig:setting-page-tsconfig>`. With this method, the user can pick predefined corporate design colors with a click instead of typing in RGB values.
```
Maybe like this?

sarahmccarthy123: Yes. I would say "select" rather than "pick" (maybe then you could say "a colorpicker" rather than "a click") but otherwise it looks fine to me.

linawolf: Done

linawolf (approved):

PR #837, opened 2023-10-27
[TASK] #102224 - Add note about removing FlexForm dataStructure lookups in v13

brotkrueml:
As the migration of the lookups can already be done in v12, a note is added to those configuration options.

Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/705
Releases: 12.4

### DIFF
--- a/Documentation/ColumnsConfig/Type/Flex/AboutDataStructures.rst
+++ b/Documentation/ColumnsConfig/Type/Flex/AboutDataStructures.rst
@@ -85,6 +85,10 @@ on a selected "tt_content" element plugin.
 Data structure lookup with tree traversal
 =========================================
 
+..  note::
+    This example will stop working with TYPO3 v13.
+    :ref:`More information <columns-flex-properties-ds-pointerfield-searchparent>`
+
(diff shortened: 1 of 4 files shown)


Documentation/ColumnsConfig/Type/Flex/AboutDataStructures.rst
```diff
@@ -85,6 +85,10 @@ on a selected "tt_content" element plugin.
 Data structure lookup with tree traversal
 =========================================
 
+..  note::
```
linawolf, line 88: I would suggest to just remove the example in v13?

brotkrueml: The PR is for 12.4 (see "Releases" line), just hinting: It will be gone in v13 but you can already migrate.

linawolf: Woops sorry, missed that part

linawolf (approved):

Issue #244, opened 2019-10-10
check slug config examples


georgringer:
see https://forge.typo3.org/issues/89390

sorry don't have currently time to dig into it, just shouldn't get lost

sypets:
there is also an open PR #244

cepheiVV:
This issue isn't complete.
The documentation states that the fallback for `nav_title // title` should be:
```
$GLOBALS['TCA']['pages']['columns']['slug']['config']['generatorOptions']['fields'] = ['nav_title', 'title'];
```

while it actually only works with this config:
```
$GLOBALS['TCA']['pages']['columns']['slug']['config']['generatorOptions']['fields'] = [['nav_title', 'title']];
```
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.5.x/Feature-87085-FallbackOptionsForSlugFields.html#examples

linawolf:
@cepheiVV where ibn the dos did you actually find this?

cepheiVV:
@linawolf 
I posted the Link to the docs in my message above

<img width="863" alt="Screenshot 2023-06-23 at 14 17 26" src="https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/assets/2963482/515c9f47-e127-438b-806e-faff92ed60fb">

linawolf:
Ah you are talking about the changelog entry https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.5.x/Feature-87085-FallbackOptionsForSlugFields.html#examples itself beeing wrong?

sypets:
As mentioned above: **Changelog** changes should be handled as core issues / patches.

The examples in the [changelog](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.5.x/Feature-87085-FallbackOptionsForSlugFields.html#examples) are wrong (example 2 and 3). I have created a patch: https://review.typo3.org/c/Packages/TYPO3.CMS/+/81723 (this should be backported to 9.5, but it is my understanding patches will not be backported to versions before 11.5 currently).


The example in the latest "main" of official docs are correct.

* "main": https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.html#confval-generatorOptions-fields

[...]

PR #854, opened 2023-11-14
Update slug generatorOptions fields

sypets:
- Explain how the array is used (in the second paragraph).
- Explain how the simple array is resolved
- If a simple array is used in the examples, also show the variant with the nested array
- In the examples, use nav_title, title if they should be used as "either or" because that makes more sense for the examples given
- Start the examples with the more simple cases of "either or", not concatenation

In the description, we assume a multi-dimensional array and treat the simple array as a special case (which can be resolved to a multi-dimensional array as well).

Resolves: #244

### DIFF
--- a/Documentation/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.rst
+++ b/Documentation/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.rst
@@ -22,7 +22,7 @@ generatorOptions
 
    In the *two-dimensional array*, elements of the first level will be concatenated,
    elements of the second level are used as several options for a slug segment,
-   where the first not empty one is used.
+   where the first one not empty is used.
 
    Thus, it can be used as nested array to combine multiple fields using both
    concatenation and several options, for example
--- a/Documentation/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.rst
+++ b/Documentation/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.rst
@@ -20,7 +20,17 @@ generatorOptions
 
    Insert several field names (of type string) that will be considered during slug construction.
 
-   Can also be used as nested array to combine multiple fields :php:`[['nav_title', 'title'], 'other_field']`.
+   In the *two-dimensional array*, elements of the first level will be concatenated,
+   elements of the second level are used as several options for a slug, where
+   the first not empty one is used.
+
+   Thus, it can be used as nested array to combine multiple fields using both
+   concatenation and several options, for example
+   :php:`[['nav_title', 'title'], 'other_field']`.
+
+   If a *simple array* is used, each string will be used in the same way, as if
+   it were enclosed in an array, for example :php:`['title', 'subtitle']` has
+   the same result as :php:`[['title'], ['subtitle']]` and results in concatenation.
 
    .. admonition:: Info
       :class: attention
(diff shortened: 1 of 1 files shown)


Documentation/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.rst
```diff
@@ -20,34 +20,42 @@ generatorOptions
 
    Insert several field names (of type string) that will be considered during slug construction.
 
-   Can also be used as nested array to combine multiple fields :php:`[['nav_title', 'title'], 'other_field']`.
+   In the *two-dimensional array*, elements of the first level will be concatenated,
```
brotkrueml, line 23: The sentence is hard to understand, at least for me. I has to read it multiple times. What is a two-dimensional array in this scope?

Documentation/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.rst
```diff
@@ -20,34 +20,42 @@ generatorOptions
...
+   elements of the second level are used as several options for a slug segment,
+   where the first not empty one is used.
```
brotkrueml, line 25:
```suggestion
   where the first one not empty is used.
```

Documentation/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.rst
sypets: > What is a two-dimensional array in this scope?

a multidimensional array with 2 levels, or for our special case: `array<array<int,string>>`

* `$a = ['a', 'b'];` is simple array with one level (1 array)
$a = [['a'], ['b']]; is multidimensional array with 2 levels ("an array of arrays") 

I don't know how else to say it.

----

So, is this a common term?

If we compare terminology with PHP docs, this is also used here:

"multidimensional array"

https://www.php.net/manual/en/language.types.array.php

Did not find "Two-dimensional array" in official PHP docs, but for example here: https://www.w3schools.com/PHP/php_arrays_multidimensional.asp

----

**But**, if is not easily understandable, maybe there is a better way to explain it.

Main intention is to find an explanation for how the values are interpreted.

What I am missing in the current version is exactly this. You can deduce it from the examples. But we have currently a pretty vague introduction:

> Insert several field names (of type string) that will be considered during slug construction.
> Can also be used as nested array to combine multiple fields [['nav_title', 'title'], 'other_field'].

It is then made clearer in the Info section, but not really enough: Why would inserting multiple fields in a simple array result in a concatenated slug? 

I first found current description confusing until I understood how it works. (I did deduce it mostly from the examples but had to look in the source code to understand behaviour of the "simple array").

---

**Alternative**: we could leave it as is and let examples speak for themselves. That is preferably, if we are only half happy with current patch.

brotkrueml: Hmm, the sentence is really hard to grasp for me. I value your intention. Also I stumbled upon "first level" and "second level" (which are not explained).

Maybe something like this:

--- start ---

A slug can consist of multiple levels, for example, the first level can be the page title and the second can be another field:

page-title-other-field

Here "page-title" is the first level and "other-field" is the second level. This is defined as ['title', 'other_field']

The first level can also be an array, like "['nav_title', 'title'], the generator option is then: [['nav_title', 'title'], 'other_field'].

.... and so on ....

 --- end ---
 
 That may be better understandable. What do you think?

sypets: I meant the level of nesting in the array. The suggestion does not convey my original intention. 

I wanted to sleep on it a bit. I will look at it again later today or weekend and hopefully we can find a solution which works.

sypets:
I created another PR https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/868 changing only the examples. I copied the examples from the changelog https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.5.x/Feature-87085-FallbackOptionsForSlugFields.html?highlight=generatoroptions

PR #855, opened 2023-11-14
Fix rendering problem due to composer.json

sypets:
The file composer.json contained a syntax
error.


sypets:
Obsolete, based on wrong version.

PR #856, opened 2023-11-14
Fix spelling of allowNonIdValues

sypets:
In some cases, this property was spelled with all-lowercase, which is not correct.


brotkrueml (approved):

brotkrueml:
@sypets Good catch! Thanks for the fix :+1:

Issue #426, opened 2021-07-30
Option "exclusiveKeys" only works for "selectMultipleSideBySide" / "selectTree"


mbrodala:
Currently the docs imply that the select option `exclusiveKeys` can be used for all render types. But this is not the case, only `selectMultipleSideBySide` and `selectTree` respect this option. All other render types do not change the selection if an exclusive key is selected. (And for `selectSingle` this would not even be possible.)

So this option should be extracted and displayed only for `selectMultipleSideBySide` / `selectTree`.

nhovratov:
Well, the DataHandler checks only for `type="select"`, so it works kinda for all render types. But you are right, that the FormEngine only validates `selectMultipleSideBySide`  and `selectTree` / `category` in the backend forms.

We should make clear, that there is this distinction.

PR #861, opened 2023-11-14
Remove exclusiveKeys property for some select renderTypes

sypets:
The property exclusiveKeys is not available for all select types, it is only available for selectMultipleSideBySide and selectTree renderType (and category type).

The property is removed for the others and a note added.

The type category is left as is.

Resolves: #426

### DIFF
--- a/Documentation/ColumnsConfig/Type/Select/CheckBox/Properties.rst
+++ b/Documentation/ColumnsConfig/Type/Select/CheckBox/Properties.rst
@@ -25,11 +25,6 @@ Select properties
 *  :ref:`items <columns-select-properties-items>`
 *  :ref:`MM <columns-select-properties-mm>`
 
-.. note::
-
-    The property :ref:`exclusiveKeys <columns-select-properties-exclusivekeys>`
-    is available only for the renderTypes selectMultipleSideBySide and selectTree.
-
 Common properties
 =================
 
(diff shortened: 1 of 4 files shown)


Documentation/ColumnsConfig/Type/Select/CheckBox/Properties.rst
```diff
@@ -26,6 +25,11 @@ Select properties
 *  :ref:`items <columns-select-properties-items>`
 *  :ref:`MM <columns-select-properties-mm>`
 
+.. note::
```
brotkrueml, line 28: I would not add a note here. We are in select->checkbox (also for the other two). But a note would be suitable on the exclusiveKeys page:
https://docs.typo3.org/m/typo3/reference-tca/12.4/en-us/ColumnsConfig/Type/Select/Properties/ExclusiveKeys.html

brotkrueml (changes requested):

Documentation/ColumnsConfig/Type/Select/CheckBox/Properties.rst
sypets: Done.

brotkrueml (approved):

brotkrueml:
@sypets Thanks again for your contribution :+1:

Issue #593, opened 2022-05-17
Duplicate link label - mm-match-fields


tomwarwick:
https://typo3.slack.com/archives/C028JEPJL/p1651130763798999

DavidBruchmann:
the essential part of the communication:

![image](https://user-images.githubusercontent.com/999146/235857876-493fe29b-94f6-49a0-83d2-4441031d406f.png)

The link inside the screenshot:
https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/blob/main/Documentation/ColumnsConfig/Type/Select/Properties/Mm.rst#L157

sypets:
Indeed, columns-select-properties-mm-match-fields is still defined twice:

```
Documentation/ColumnsConfig/Type/Select/Properties/Mm.rst:..  _columns-select-properties-mm-match-fields:
Documentation/ColumnsConfig/Type/Select/Single/Properties.rst:.. _columns-select-properties-mm-match-fields:

```

Will create PR.

PR #863, opened 2023-11-14
Remove duplicate label definition

sypets:
_columns-select-properties-mm-match-fields: should not exist twice.

Resolves: #593


brotkrueml (approved):

brotkrueml:
@sypets Thanks for your work :+1:

PR #868, opened 2023-11-18
Change slug generatorOptions.fields examples

sypets:
Reuse the same examples from the changelog with
the following changes:

- For nav_title, title instead of concatenate we use "or" (use nav_title if not empty, else use title) This makes more sense in a realworld example.
- In the second columns, set only the values for the fields which are used in the first column
- For the example of array of strings, add the notation for array as array as well (example 3)


Documentation/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.rst
```diff
@@ -33,21 +33,19 @@ generatorOptions
...
+   +-----------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------+-----------------------------------+
+   |:php:`['nav_title', 'title'], 'subtitle'`                        | :php:`['title' => 'Products', 'nav_title' => 'Best products', 'subtitle' => 'Product subtitle']`                                     | `/best-products/product-subtitle` |
+   +-----------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------+-----------------------------------+
+   |:php:`['seo_title', 'title'], ['nav_title', 'subtitle']`         | :php:`['title' => 'Products', 'nav_title' => 'Best products', 'subtitle' => 'Product subtitle', 'seo_title' => 'SEO product title']` | `/seo-product-title/products`     |
```
brotkrueml, line 47:
```suggestion
   |:php:`['seo_title', 'title'], ['nav_title', 'subtitle']`         | :php:`['title' => 'Products', 'nav_title' => 'Best products', 'subtitle' => 'Product subtitle', 'seo_title' => 'SEO product title']` | `/seo-product-title/best-products`|
```

brotkrueml (changes requested):

brotkrueml (approved):

brotkrueml:
@sypets Thanks for your work :+1:

Issue #663, opened 2022-11-01
Improve description and examples for foreign_table_where (selectTree)


sypets:
https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.html#foreign-table-where

## Problem

The `foreign_table_where`  property is actually an "append existing WHERE clause (starting with AND) and optionally add more such as ORDER BY / GROUP BY /  LIMIT" clause. This is made clear in the example below, but not in the description.

> Type string (SQL WHERE)
> The items from [foreign_table](https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/Select/Properties/ForeignTable.html#columns-select-properties-foreign-table) are selected with this WHERE-clause.

https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.html#foreign-table-where

## Suggestions:

* update the short description in the block at the top in [foreign_table_where](https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.html#columns-select-properties-foreign-table-where) page to use something different than "WHERE clause" which might make clear you can also add only an "ORDER BY" (as used in the "happy little tree" example etc.), only WHERE or combination, etc. -  ("ORDER BY", "LIMIT" etc. are technically not part of a where clause). (This is explained in the description below, but the short description at the top is misleading)

PR #870, opened 2023-11-18
Clarify foreign_table_where

sypets:
The previous explanation for foreign_table_where was slightly misleading. The examples speak for themselves, but they are further down on the page.

The WHERE-clause is effectively appended to the existing WHERE clause created by TYPO3 using the default values and must start with AND.

Also, additional clauses such as ORDER BY or GROUP BY may be added after the WHERE-clause.

Resolves: #663

### DIFF
--- a/Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
+++ b/Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
@@ -13,7 +13,11 @@ foreign_table_where
    :RenderType: all
 
    The items from :ref:`foreign_table <columns-select-properties-foreign-table>`
-   are selected with this WHERE-clause.
+   are selected with this WHERE-clause. The WHERE-clause is effectively
+   appended to the existing WHERE-clause (which contains default constraints,
+   such as 'NOT deleted') and must begin with AND. Additional clauses such as
+   ORDER BY may be added after the WHERE-clause. The result must be a valid SQL
+   statement.
 
 Field quoting
 =============
--- a/Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
+++ b/Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
@@ -13,11 +13,9 @@ foreign_table_where
    :RenderType: all
 
    The items from :ref:`foreign_table <columns-select-properties-foreign-table>`
-   are selected with this WHERE-clause. The WHERE-clause is effectively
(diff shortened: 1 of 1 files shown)


Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
```diff
@@ -13,7 +13,11 @@ foreign_table_where
    :RenderType: all
 
    The items from :ref:`foreign_table <columns-select-properties-foreign-table>`
-   are selected with this WHERE-clause.
+   are selected with this WHERE-clause. The WHERE-clause is effectively
```
brotkrueml, line 16: "WHERE clause" please without dash. Also I suggest to embed WHERE in the text role: :sql:

Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
```diff
@@ -13,7 +13,11 @@ foreign_table_where
...
+   appended to the existing WHERE-clause (which contains default constraints,
+   such as 'NOT deleted') and must begin with AND. Additional clauses such as
+   ORDER BY may be added after the WHERE-clause. The result must be a valid SQL
```
brotkrueml, line 19: Which result? This can be misleading as the content of the where clause is not valid SQL.

Maybe something like "The statement to be executed must be valid SQL". But that is obvious.

Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
sypets: ok, but this is how it is currently on the page.

sypets: done.

Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
sypets: Removed ambiguous sentence.

Documentation/ColumnsConfig/Type/Select/Properties/ForeignTableWhere.rst
brotkrueml: Yup, I know :-( - but we should make it better if we rework a section.

brotkrueml (approved):

brotkrueml:
@sypets Thanks again for the improvement :+1:

PR #893, opened 2023-12-21
Fix missing backtick

dogawaf:
Can be backported in 12.4 and 11.5 where the typo also exists.


brotkrueml (approved):

brotkrueml:
@dogawaf Good catch! Thanks for the contribution :+1:

PR #897, opened 2023-12-29
Fix guilabel

simonschaufi:


simonschaufi:
this is fixed in main already but was not backported.

brotkrueml (approved):

PR #899, opened 2024-01-10
Update allowed image types

misterboe:
['jpg','png','svg'] causes: PHP Warning: Array to string conversion in /var/www/html/vendor/typo3/cms-backend/Classes/Form/FieldControl/ElementBrowser.php line 77

PR #902, opened 2024-01-11
Update allowed image types (#899)

brotkrueml:
['jpg','png','svg'] causes: PHP Warning: Array to string conversion in /var/www/html/vendor/typo3/cms-backend/Classes/Form/FieldControl/ElementBrowser.php line 77


brotkrueml:
Forward-port of #899

linawolf (approved):

PR #903, opened 2024-01-11
[TASK] Use PHP 8.2 and allow Symfony 7

brotkrueml:
Also switch to php-cs-fixer directly, as typo3/coding-standards are not compatible with Symfony 7 by now. The fixer rules from Core are used.

Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/781
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/786
Releases: main

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,8 +1,74 @@
 <?php
 
-$config = \TYPO3\CodingStandards\CsFixerConfig::create();
-$config
-    ->getFinder()->in(__DIR__)
-;
+declare(strict_types=1);
 
-return $config;
+use PhpCsFixer\Config;
+use PhpCsFixer\Finder;
+
+return (new Config())
+    ->setFinder(
+        (new Finder())
+            ->in(__DIR__)
+    )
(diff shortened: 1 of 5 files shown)


.php-cs-fixer.dist.php
```diff
@@ -1,8 +1,74 @@
...
+use PhpCsFixer\Config;
+use PhpCsFixer\Finder;
+
+return (new Config())
+    ->setFinder(
+        (new Finder())
+            ->in(__DIR__)
+    )
+    ->setRiskyAllowed(true)
+    ->setRules([
```
linawolf, line 14: Where do all these extra rules come from?

brotkrueml: As written in the commit message, I am using the rules from Core:
https://github.com/TYPO3/typo3/blob/main/Build/php-cs-fixer/config.php

But we can stick with PER-CS@1 and small adjustments like in render-guides.

linawolf: ah ok, Ic

linawolf (approved):

PR #910, opened 2024-01-13
[TASK] Add missing itemGroups property to category type

brotkrueml:
Additionally, "RenderType" for properties "foreign_table" and "foreign_table_prefix" have been removed as category has no render types. The toctree is also ordered alphabetically.

Releases: main, 12.4

### DIFF
--- a/Documentation/ColumnsConfig/Type/Category/Properties/Index.rst
+++ b/Documentation/ColumnsConfig/Type/Category/Properties/Index.rst
@@ -1,6 +1,6 @@
-.. include:: /Includes.rst.txt
+..  include:: /Includes.rst.txt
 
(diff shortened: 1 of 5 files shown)


Documentation/ColumnsConfig/Type/Category/Properties/Index.rst
```diff
@@ -13,12 +13,13 @@ Special properties
```
linawolf, line 13: Maybe use a glob, if all properties should be listed alphabetically

brotkrueml: Ack

linawolf (approved):

PR #912, opened 2024-01-16
Fix nullable example

misterboe:
nullable only works with true/false not with single quotes


brotkrueml (approved):

brotkrueml:
@misterboe Good catch. Thanks for the fix :+1: 

Please make PRs against the main branch in the future, this eases the backporting.

misterboe:
This seems to happen when using the "edit on github" button in the docs. But will check that next time

PR #915, opened 2024-01-17
Fix nullable example (#912)

brotkrueml:
nullable only works with true/false not with single quotes


brotkrueml:
Forward-port from https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/912

linawolf (approved):

PR #916, opened 2024-01-17
[TASK] Remove wrong default for type `file` `allowed`

nhovratov:
This is not documented anywhere and didn't work as well when testing. The code for this is:

```
if (!empty(($allowed = ($fieldConfig['config']['allowed'] ?? null)))) {
    $fieldConfig['config']['allowed'] = self::prepareFileExtensions($allowed);
}
```

Releases: main, 12.4


brotkrueml (approved):

brotkrueml:
@nhovratov Thanks for the improvement :+1:

PR #929, opened 2024-01-31
Update list of valid searchable field types

nhovratov:
Added uuid and json as valid field types. These were added in TYPO3 v12 and are available to be used in the live search.


nhovratov:
For reference:

https://github.com/TYPO3/typo3/blob/2e0fd8c05d62e6187baeb806d5bd59018c45e842/typo3/sysext/backend/Classes/Search/LiveSearch/DatabaseRecordProvider.php#L443-L458

brotkrueml (approved):

brotkrueml:
@nhovratov Thanks for adding the missing parts :+1:

PR #931, opened 2024-02-02
[TASK] Fix wrong default renderType for type check

nhovratov:
The default value for `renderType` is always equivalent to the `type`. There does not exist a renderType with the value `default`. The actual default is the value `check`, the same as the type.

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ColumnsConfig/Type/Check/Default.rst
+++ b/Documentation/ColumnsConfig/Type/Check/Default.rst
@@ -8,10 +8,10 @@
 Default checkbox
 ================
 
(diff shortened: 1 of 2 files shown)


nhovratov:
For reference in the Core: https://github.com/TYPO3/typo3/blob/10585f2dce13770fc73043e3d319b6f7cc796f46/typo3/sysext/backend/Classes/Form/Container/SingleFieldContainer.php#L122-L131

garvinhicking (approved):
Thank you for this fix, that "default" fallback is described clearer now! :)

PR #934, opened 2024-02-10
Add hint of removal in TYPO3 12

franzholz:


franzholz:
I consider it useful. If you write a new PHP code then you should be aware if something will soon not be there any more.

brotkrueml:
We have many of that changes where properties vanish in newer versions or functionality changes. So you propose we should add them all to older docs? That would be the consequence if we add one. Why not all the others? What is so special about cruser_id that it should be an exception?

franzholz:
I do not say that all of them need to be added. But if someone notices something, then it should be added.
cruser_id has been with TYPO3 since the beginning and it has been used very much. It is very helpful if users get informed about this.

sbuerk:
> It is very helpful if users get informed about this.

Ack. But they are. That's exactly why they are noticed in the changelog. To inform people upgradeing to a major version.

Changelog states what changes exactly between versions and how to deal with them. Therefore, as long as the Documentation Team removes removed fields in that version it is removed everything is fine. The documentation should show the state of the exact version. For version differences the changelogs are the source of thruth.

Otherwise, the Documentation Team would have even more personal issues and they are doing a great job.

I'd vote against adding this here as Chris mentioned. Starting with one, they will have to do that for every change and even prior versions.

linawolf:
I agree with @brotkrueml and @sbuerk, this information should be sufficient by being mentioned in the changelog at it would overwhelm us as a team to mention each future change in back versions.

PR #944, opened 2024-03-07
[BUGFIX] Remove "showRemovedLocalizationRecords" from "inline" example (v10)

mbrodala:
This option was removed with TYPO3v7, see https://review.typo3.org/c/Packages/TYPO3.CMS/+/44363


brotkrueml (approved):

brotkrueml:
@mbrodala Thanks for taking care :+1:

PR #956, opened 2024-03-07
remove field "t3ver_id" from docs

can0199:
in typo3 version 10 this field was removed.

Releated Changelog:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.0/Breaking-88143-Version-relatedDatabaseFieldT3ver_idRemoved.html


brotkrueml (approved):

brotkrueml:
@can0199 Thanks for spotting this :+1:

PR #964, opened 2024-03-10
[TASK] Switch codeblock to sql

linawolf:


brotkrueml (approved):
This may also vanish because it is created automatically, if a table is workspace-aware.

PR #975, opened 2024-03-15
nullable is still missing in the docs

franzholz:
See TYPO3 12.0 

Feature: #97384 - TCA option "nullable"


brotkrueml (commented):
"nullable" is no common property. For example, the TCA type "file" cannot be nullable:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-98479-NewTCATypeFile.html

This property should be mentioned at the according TCA types. Just looked at "email", where it is available: https://docs.typo3.org/m/typo3/reference-tca/12.4/en-us/ColumnsConfig/Type/Email/Index.html

PR #981, opened 2024-03-23
[BUGFIX] Resolve rendering warnings

linawolf:
releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ColumnsConfig/Type/Inline/Properties/ForeignTable.rst
+++ b/Documentation/ColumnsConfig/Type/Inline/Properties/ForeignTable.rst
@@ -1,15 +1,15 @@
-.. include:: /Includes.rst.txt
-.. _columns-inline-properties-foreign-table:
+..  include:: /Includes.rst.txt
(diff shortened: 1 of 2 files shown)


brotkrueml:
Which warnings?

linawolf:
If you run the tests on the current situation there are two warnings about duplicate confvals

brotkrueml:
So, the other confvals for foreign-table and title should also get a name to be consistent?

linawolf:
This PR is about resolving the warnings, so that the tests are green again. there are many confvals so renaming them all consistently is a task beyond the scope of this PR

brotkrueml:
Personally, I would add all names, which then also removes the warnings (as drive-by). This way now, it is half-baken: you have no warnings anymore, but inconsistent confval configurations - which must be addressed in one or more separate PRs (and not forgotten to do so).

brotkrueml (approved):
It may be okay for now. But add the other names in near time as separate PRs to be consistent.

PR #982, opened 2024-03-25
[TASK] Unify confval names in ctrl section

linawolf:
releases: main, 12.4, 11.5


brotkrueml (approved):
Personally, I find the mix of lowercase (name, type) and uppercase (Path, Scope) irritating. Is there a rule?

linawolf:
yes: there are defined properties (type, name, noindex, .. those are in lowercase. All properties that are not predefined are used as labels and will be printed exactly the way they are written https://sphinx-toolbox.readthedocs.io/en/stable/extensions/confval.html

PR #1026, opened 2024-03-28
[TASK] Overhaul common pages indentation

linawolf:
releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/Introduction/Index.rst
+++ b/Documentation/Introduction/Index.rst
@@ -87,21 +87,15 @@ The "first level" of the :php:`$GLOBALS['TCA']` array is made of the table names
 they appear in the database):
 
 ..  code-block::php
-    :caption: EXT:my_extension/Configuration/Overrides/pages.php
+    :caption: The array looks like this
 
     $GLOBALS['TCA']['pages'] = [
         // ...
     ];
 
-..  code-block::php
-    :caption: EXT:my_extension/Configuration/Overrides/tt_content.php
-
     $GLOBALS['TCA']['tt_content'] = [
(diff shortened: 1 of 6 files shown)


Documentation/Introduction/Index.rst
```diff
@@ -111,25 +119,29 @@ Each table is further defined by an array which configures how the
...
-       ],
-       'palettes' => [
-           ....
-       ],
-   ];
+The general structure (looking at a single table) is as follows:
+
+
+..  code-block::php
+    :caption: EXT:my_extension/Configuration/Overrides/tx_examples_haiku.php
```
brotkrueml, line 126: I am not sure if "Overrides" is the correct path here: In the example you override the whole TCA for tx_examples_haiku - which is not recommended in Overrides, I assume. This is more a "how is the TCA built", not "how you should override and place it", so I suggest to remove the caption altogether.

Applies also above for "Table entries (first level)"

linawolf: Mhh, the syntax would not be correct directly in the TCA folder either. What would you recommend as a caption to make this clear?

linawolf: Like this maybe?

brotkrueml: Does the caption "The array looks like this" really helps? I would remove the caption, we don't need it IMHO.

linawolf: Then you cannot know if the information where to put this is only missing or if this not an example you should use

brotkrueml (approved):

PR #1032, opened 2024-04-08
[TASK] Improve array/string notation for TCA 'allowed' on file types

garvinhicking:
After an initiative of  Lars Trebing (via https://typo3.slack.com/archives/C03AM9R17/p1712352682709819) we need to clarify this part of the documentation.

It is not explained more thoroughly how string | array definitions can be used, and that it only works with an Array starting with the next TYPO3 release (likely tomorrow). The relevant bugfix for this is https://review.typo3.org/c/Packages/TYPO3.CMS/+/83649.

For reference, the changelog https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-98479-NewTCATypeFile.html explicitly allows:

> It's possible to use **multiple placeholders**. 
> It's also possible to **mix them** with single file extensions. 
> Additionally, it's also possible to define the **file extensions as `array`**.

(I ensured the examples I listed work in current TYPO3 main+12.4-dev)

Further references on previous changes:
https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/899
https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/902

Releases: main, 12.4

### DIFF
--- a/Documentation/ColumnsConfig/Type/File/Properties/Allowed.rst
+++ b/Documentation/ColumnsConfig/Type/File/Properties/Allowed.rst
@@ -11,7 +11,7 @@ allowed
     :type: string / array
     :Scope: Proc. / Display
 
-    One of the following reserved strings:
+    One or more (comma separated) of the following reserved strings:
 
     `common-image-types`
         Gets replaced with the value from
@@ -25,5 +25,15 @@ allowed
         Gets replaced with the value from
         :php:`$GLOBALS['TYPO3_CONF_VARS']['SYS']['mediafile_ext']`.
 
-    or an array of allowed file
-    endings, for example :php:`['jpg','png','svg']`.
+    Additionally, specific allowed file extensions can be added (comma
+    separated), for example
+    :php:`'common-image-types, common-text-types, gz, zip`.
+
+    You can also use the array notation of allowed file extensions, for example
+    :php:`['jpg','png','svg']` or :php:`['common-image-types', 'gz', 'zip']`.
+
+..  versionadded:: 12.4.14
+    Due to a bug, the array notation only properly works
+    since TYPO3 v12.4.1 (and 13.1.0) and upwards. Use the string notation
+    for earlier versions instead.
+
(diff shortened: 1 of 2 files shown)


garvinhicking:
(Note: We should only merge this after the T3v12 release has been made tomorrow to not cause confusion)

linawolf:
Converting this to draft so it does not accidently get merged eatly

garvinhicking:
@linawolf Thanks! Release was made, should we merge this?

linawolf (approved):

linawolf:
Sure, please go ahead

Documentation/ColumnsConfig/Type/File/Properties/Allowed.rst
```diff
@@ -11,7 +11,7 @@ allowed
     :type: string / array
     :Scope: Proc. / Display
 
-    One of the following reserved strings:
+    One or more (comma separated) of the following reserved strings:
```
brotkrueml, line 14:
```suggestion
    One or more (comma-separated) of the following reserved strings:
```

brotkrueml (commented):
Hmm, actually, what is the recommendation? using comma-separated or array?

garvinhicking:
> Hmm, actually, what is the recommendation? using comma-separated or array?

Well, I feel like the Core team would like to advocate for the array syntax, but that's a compatibility issue.

What do we usually "recommend"? Something that is forward compatible, or something that works now and for the most cases?

To play it safe, we should recommend string syntax. To set ourselves up for the future, we should recommend array syntax.

(That's why I didn't really want to make a strong recommendation but let it up for the reader).

I can live with all three ways (no recommendation, recommend array, recommend string), so feel free to adapt this commit and close/merge.

brotkrueml:
Hmm, as this works mainly beginning with v12.4.14, I think, we leave it as for now. Our future selfs (or other people) should decide later.

brotkrueml (approved):

PR #1049, opened 2024-05-13
prefixParentPageSlug is only for pages

franzholz:
prefixParentPageSlug should not be mentioned in record examples.

https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/Slug/Properties/GeneratorOptions.html#generatoroptions

Do not immediately accept this pull request.
Tell me your opinion, then I will add more example files where the prefixParentPageSlug should be removed.

### DIFF
--- a/Documentation/CodeSnippets/Slug5.rst.txt
+++ b/Documentation/CodeSnippets/Slug5.rst.txt
@@ -17,7 +17,6 @@
                                'input_2',
                            ],
                        ],
-                       'prefixParentPageSlug' => false,
                    ],
                    'fallbackCharacter' => '-',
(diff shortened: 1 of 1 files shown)


linawolf:
The file you changed is an auto-generated code-snipped. The original code comes from the extension typo3/cms-styleguide. So please propose your changes there.

franzholz:
Thanks for the information.

I have now created a pull request there:

https://github.com/TYPO3-CMS/styleguide/pull/481

linawolf:
Once https://github.com/TYPO3-CMS/styleguide/pull/481 the next update will automaticlaly change the codesnippets here. We do not change them manually.

PR #1055, opened 2024-05-30
Example Added for the use of user (UserFunc)

H-OsamaSultan:
Corrected the documentation for the placement of UserFunc also added example how to use it.


linawolf (approved):
Thank you for your contribution!

PR #1069, opened 2024-06-21
Update DisplayCond.rst

metapublic-gbr:


froemken (approved):
Checked that in source code of current TYPO3 version. This condition is still valid. Can someone of doc team check, if we can backport that PR to TYPO3 10|11|12? Thank you for your contribution.

Documentation/Columns/Properties/DisplayCond.rst
```diff
@@ -207,3 +207,5 @@ Flex form fields can access field values from various different sources:
     <displayCond>FIELD:flexField_1:!=:foo</displayCond>
     <!-- Hide field if value of field "flexField_1" from sheet "sheet_1" is not "foo" -->
     <displayCond>FIELD:sheet_1.flexField_1:!=:foo</displayCond>
+    <!-- Hide field for new records -->
+    <displayCond>REC:NEW:true</displayCond>
```
mbrodala, line 211: This is not restricted to FlexForms and especially does not fit the purpose of this doc section. This section explains the ways to access field values given the special nature of FlexForms.

Instead this must be moved to the general examples above, then with regular PHP/TCA syntax. Also a 2nd example for `REC:NEW:false` should be added for clarity.

mbrodala (changes requested):

Documentation/Columns/Properties/DisplayCond.rst
froemken: Good point

linawolf: Can we do that in a follow up As the current example is not wrong?

mbrodala: The example itself is wrong but it is wrongly placed, thus unexpected and won't be discovered by people. So adding it like this would not add as much value as it could.

mbrodala: I just updated the change and moved everything upwards.

mbrodala (approved):

linawolf (approved):

linawolf:
Thanks a lot for your contribution @metapublic-gbr and for moving the info @mbrodala

PR #1070, opened 2024-06-24
Fix incorrect field name in 'translationSource'

creifenscheid:
Replaces "l10n_parent" with "l10n_source" in the convention hint of 'ctrl'-property 'translationSource'


froemken (approved):

froemken:
Thank you for providing this patch.

PR #1086, opened 2024-07-12
Consistent use of l10n_ as prefix for all localization fields

ghermens:
As mentioned several times throughout the TCA documentation, l10n_ should be used by convention, and l18n_ must be kept for some core tables for historic reasons. l10n_ should therefore be used consistently, except when explaining those historic reasons.


linawolf (approved):
Thanks for catching these!

PR #1099, opened 2024-07-14
Add info about wrong icon usage

froemken:
It's the third time now that I stumbled over the problem, that in create new record wizard the icon of my group is from a foreign extension. This time it was the icon of tx_scheduler, which has nothing to do with my extension. TYPO3 tries to find a loaded extension by the second part of the tablename. This is OK for all extensions without underscore. But with underscore the second part of table name does not match any extension key. That's why it comes to these side-effects. Easiest solution is to enter a localized title. In that case the extension key will be extracted from the LLL:EXT: path. Else you should set groupName to your extension key. That way an extension can be found and the correct group icon (ext icon) will be shown again

### DIFF
--- a/Documentation/Ctrl/Properties/GroupName.rst
+++ b/Documentation/Ctrl/Properties/GroupName.rst
@@ -5,13 +5,21 @@
 groupName
 =========
 
-.. confval:: groupName
-   :name: ctrl-groupName
-   :Path: $GLOBALS['TCA'][$table]['ctrl']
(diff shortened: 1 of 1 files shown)


froemken:
For TYPO3 6.2 I have added that forge ticket: https://forge.typo3.org/issues/57102
Problem with title is solved, but group still contains wrong icon

linawolf (approved):

PR #1133, opened 2024-08-02
Update Index.rst, fixed spelling

mtness:
Updated spelling, see
https://www.scribbr.com/us-vs-uk/practise-or-practice/ for details.

"to practise is only a verb in UK, in US it is never used"


brotkrueml (approved):
@mtness Thanks for the fix :+1:

PR #1159, opened 2024-08-29
[TASK] Document `hideDeleteIcon` for TCA "Group fields"

sascha-egerer:
releases: main, 12.4, 11.5


linawolf (approved):

linawolf:
Thanks for this change, seems like it has been missing ever since.

An automatic backport to TYPO3 v11 is not possible as the structure of the TCA reference has changed since. If you desire to backport this to 11.5 you would need to make a separate PR based on branch 11.4

PR #1182, opened 2024-10-12
[TASK] Deprecate TCA subtypes

linawolf:
References: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1073
Releases: main

PR #1183, opened 2024-10-12
[TASK] Deprecate plugin content element and plugin subtypes (list_type)

linawolf:
Remove further mentioning of the list_type not covered by #1182

Reference: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1071
Releases: main


sfroemkenjw (approved):
`list_type` is deprecated and may be removed with TYPO3 14. So, having `list_type` and `CType` in one example is not needed anymore. Sure, having multiple columns would still be possible, but that is a very very rare edge-case. So, removing that section is OK for me.

PR #1184, opened 2024-10-16
Fix Wizard Registration example

Smove:


linawolf (approved):

linawolf:
Thanks for fixing this

PR #1223, opened 2025-02-06
fix: format code block for none type correctly

brotkrueml:
Releases: main, 13.4, 12.4


brotkrueml:
See: https://docs.typo3.org/m/typo3/reference-tca/main/en-us/ColumnsConfig/Type/None/Index.html#columns-none-properties-fieldinformation

linawolf (approved):

PR #1249, opened 2025-04-07
Set label/value as required in ValuePicker

froemken:


linawolf (changes requested):
I already asked @georgringer the same: I find neither a breaking change nor a deprecation entry about this beeing required now. If the old way was deprecated we would have to make a deprecation notice. We cnanot require new TCA settings without a path of deprecation or migration.

georgringer:
Looks good but tbh it is more important to be aligned with all other docs. Fully trust you

linawolf:
@froemken I would write recommended then. and mention the automatic migration. Cause if it was enforced it would be troublesome to extension authors supporting two versions and we would have to write something about that.

froemken:
I updated the patch

linawolf (approved):

PR #1271, opened 2025-05-21
Update hint about TCA:fieldInformation

froemken:
The description in docs CoreAPI tells something different as written here. It say that all these three extending features are implemented for nearly all form fields. I have updated that part here and repaired the defect :ref: link

### DIFF
--- a/Documentation/ColumnsConfig/CommonProperties/FieldInformation.rst
+++ b/Documentation/ColumnsConfig/CommonProperties/FieldInformation.rst
@@ -22,8 +22,10 @@ fieldInformation
 
 ..  hint::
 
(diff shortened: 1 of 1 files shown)


froemken:
Can not find "tcaDescription" in ChangeLog of TYPO3 code.

Just checked. Following Elements does not contain "tcaDescription":

- AbstractFormElement
- InputHiddenElement
- MfaInfoElement
- PasswordElement
- UnknownElement
- UserElement

linawolf (approved):

linawolf:
Makes sense as these fields are not commonly displayed / password fileds get a description from the policy

PR #1276, opened 2025-05-24
Update section fieldInformation

froemken:
Section fieldControl and fieldWizard also have their own sections. I thought it would be good to give fieldInformation also its own section. I have added images to describe each field information much better and added a note about which HTML tags are allowed within this section. Also all form elements supporting tcaDescription are added.


Documentation/ColumnsConfig/CommonProperties/FieldInformation/Index.rst
```diff
@@ -5,27 +5,24 @@
...
-    except in cases where it does not make sense. You can use
-    :ref:`description <columns-properties-description>` to display general
-    information between a fields title and the form element itself.
+
+Currently, TYPO3 comes with following implemented `fieldInformation` nodes:
+
+..  toctree::
+    AdminIsSystemMaintainer
+    BackendLayoutFromParentPage
+    TcaDescription
```
linawolf, line 23:
```suggestion

..  toctree::
    :titlesonly:

    AdminIsSystemMaintainer
    BackendLayoutFromParentPage
    TcaDescription
```
I would use titlesonly so that subheaders are hidden here as we only want to list the config options

Documentation/CodeSnippets/Manual/FieldInformation.rst.txt
```diff
@@ -0,0 +1,22 @@
+..  code-block:: php
```
linawolf, line 1: I would suggest to move this to a php file instead and include it via literalinclude. This way the content is linted and code styles are applied: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/Reference/ReStructuredText/Code/Codeblocks.html#literalinclude

Documentation/ColumnsConfig/CommonProperties/FieldInformation/AdminIsSystemMaintainer.rst
```diff
@@ -0,0 +1,50 @@
+..  include:: /Includes.rst.txt
+..  _tca_property_fieldInformation_adminIsSystemMaintainer:
+
+=======================
+adminIsSystemMaintainer
```
linawolf, line 5: Not really sure this needs documentation, it is only used by the Core and more an internal option. Can you think of any use case this can or should be used?

Documentation/ColumnsConfig/CommonProperties/FieldInformation/BackendLayoutFromParentPage.rst
```diff
@@ -0,0 +1,47 @@
+..  include:: /Includes.rst.txt
+..  _tca_property_fieldInformation_backendLayoutFromParentPage:
+
+===========================
+backendLayoutFromParentPage
```
linawolf, line 5: This also seems to be an internal option only used by the Core. It has never been mentioned in a Changelog or something like that so I would argue that it is probably internal

Documentation/ColumnsConfig/CommonProperties/FieldInformation/TcaDescription.rst
```diff
@@ -0,0 +1,105 @@
...
+
+    ..  note::
+        The HTML tags for this specific area are limited to:
+        `<a>`, `<br>`, `<br/>`, `<div>`, `<em>`, `<i>`, `<p>`, `<strong>`,
+        `<span>`, `<code>`.
+
+`tcaDescription` is activated for most of the TYPO3 form elements by default.
+Currently, following form elements supports `tcaDescription`:
+
+*   BackendLayoutWizardElement.php
```
linawolf, line 29: Listing the PHP files which implement the elements in the core is probably not very helpful to developers? We could list all field config names but I would probably ommit it as keeping such a list up to date would be a challenge...

Documentation/ColumnsConfig/CommonProperties/FieldInformation/Index.rst
froemken: Done

Documentation/ColumnsConfig/CommonProperties/FieldInformation/AdminIsSystemMaintainer.rst
froemken: You're right. Should not be part of TCA docs

Documentation/ColumnsConfig/CommonProperties/FieldInformation/TcaDescription.rst
froemken: I understand the afford to keep such lists up2date. Removed.

Documentation/ColumnsConfig/CommonProperties/FieldInformation/TcaDescription.rst
```diff
@@ -0,0 +1,55 @@
...
+
+..  _tca_property_fieldInformation_tcaDescription_examples_activateTcaDescription:
+
+Activate tcaDescription
+-----------------------
+
+For most of the TYPO3 own form elements it is not needed to activate
+`tcaDescription` as it is activated by default. Before adding `tcaDescription`
+to any kind of form elements you have to make sure the element supports the
+rendering of `fieldInformation`.
```
linawolf, line 41:
```suggestion
rendering of `fieldInformation  <https://docs.typo3.org/permalink/t3tca:confval-fieldinformation>`_.
```
I would link that for easier navigation

PR #1283, opened 2025-06-16
[TASK] Advertise DI in creating custom FormElements

garvinhicking:
See https://forge.typo3.org/issues/106875#change-544592

The current docs do not mention that DI needs to be used. This is now added.


garvinhicking:
> thank you for that patch, but there are too many confusing things around in your patch.

Did you read the forge ticket? And the changelog? The changelog explicitly mentions services.yaml. And the reporter fixed his issue by using this.

garvinhicking:
Closing this, superseded by the better approach in https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TCA/pull/1284 - thanks Stefan!

PR #1381, opened 2026-04-22
[TASK] Clarify rows/cols intent for TCA RTE fields

garvinhicking:
The TCA attribute is evaluated only for the initial HTML DOM rendering, but later will be re-styled by the RTE integration.

Related: https://forge.typo3.org/issues/109677
Releases: main, 14.3, 13.4

### DIFF
--- a/Documentation/ColumnsConfig/Type/Text/Default/_Properties/_Cols.rst.txt
+++ b/Documentation/ColumnsConfig/Type/Text/Default/_Properties/_Cols.rst.txt
@@ -9,7 +9,8 @@
     Abstract value for the width of the :code:`<textarea>` field. To set
     the textarea to the full width of the form area, use the value 50. Default is 30.
 
-    Does not apply to RTE fields.
+    Does not apply to RTE fields, is only applied to their initial
+    unrendered state.
 
 ..  rubric:: Example: A simple text editor with 2 rows
 
--- a/Documentation/ColumnsConfig/Type/Text/Default/_Properties/_Rows.rst.txt
+++ b/Documentation/ColumnsConfig/Type/Text/Default/_Properties/_Rows.rst.txt
@@ -13,7 +13,8 @@
     automatically be increased if the content in the field is found to be of a certain length, thus the
     field will automatically fit the content. Default is 5. Max value is 20.
 
-    Does not apply to RTE fields.
+    Does not apply to RTE fields, is only applied to their initial
+    unrendered state.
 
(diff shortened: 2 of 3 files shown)


stepo2:
Thanks, but I think this new text is still not correct. Even with the CKEeditor overlay, it has the amount of rows as configured (with `'rows' => 1` it will dynamically get bigger when hitting enter for a newline, however). So this setting actually affects non-RTE exactly as RTE fields (just that RTE-text has the nice GUI feature of dynamically getting bigger if needed).

If this is considered a doc bug I think this doc line has to vanish completly, because `rows` work the same for RTE and non-RTE.

garvinhicking:
I was under the impression that CSS of RTE sets a min height here?

froemken (commented):
Hi @garvinhicking,

thank you for taking the time to look into this and opening the PR!

However, after checking the Core implementation in `\TYPO3\CMS\RteCKEditor\Form\Element\RichTextElement::render()`, the proposed clarification is technically not accurate:

In `RichTextElement::render()`, the `<textarea slot="textarea">` is rendered with a hard-coded value:
```php
$textareaAttributes = GeneralUtility::implodeAttributes([
    'slot' => 'textarea',
    'id' => $fieldId,
    'name' => $itemFormElementName,
    'rows' => '18',
    'class' => 'form-control',
    ...
], true);
```

Neither `$config['rows']` nor `$config['cols']` from the TCA definition are evaluated or passed to the textarea in `RichTextElement`. As a result, the `<textarea>` in its initial unrendered DOM state always has `rows="18"`, regardless of what was configured in TCA.

Once CKEditor 5 initializes, the dimensions are governed strictly by the RTE configuration (YAML profile / CSS rules like `.ck-editor__editable`), completely bypassing TCA.

Therefore, the current statement:
> *Does not apply to RTE fields.*

is factually correct. Adding that it applies to the unrendered state would be misleading for integrators, as setting `rows` or `cols` in TCA has zero effect on RTE fields at any stage of rendering.

If we want to improve the documentation for developers looking to adjust the dimensions of rich text fields, we could instead add a note linking to the RTE configuration, for example:
> *Does not apply to RTE fields. To adjust the height or styling of rich text fields, configure your CKEditor YAML preset or apply custom CSS (see [RTE CKEditor documentation](https://docs.typo3.org/c/typo3/cms-rte-ckeditor/main/en-us/)).*

Unless there is another place where this is evaluated that we missed, we recommend closing this PR or turning it into a cross-reference hint to the RTE configuration. What do you think?

PR #1394, opened 2026-05-19
[TASK] Clarify usage of extension configuration settings in _ForeignTableWhere.rst.txt, section "###SITE:<KEY>.<SUBKEY>###"

mz-aimcom:
Add a hint for the usage of extension configuration settings with a short description/example.


linawolf:
Thank you for your contribution. This sound like a new feature? Do you know when and with what changelog entry it was introduced?

mz-aimcom:
@linawolf You're welcome!

I don't think this is a new feature. As the setting is regularly available via site settings. At least, I expected it to work after reading the documentation and trying to use it that way.

I only added the clarification/example to the documentation because I had a bit of trouble figuring out how to set and get the variable correctly. I think this use case is very common, so it might help others succeed a little faster than I did. :)

linawolf (approved):

PR #1448, opened 2026-09-12
[TASK] Add the missing captions to 18 literalinclude blocks

linawolf:
The snippet name a literalinclude shows — _basic.php,
_multipleCropVariants.php — says what the example is about, not where it
goes, and in this manual that is the question with two answers: a table
of your own or an override of an existing one.

The captions follow the shape of each snippet. Four return a whole TCA
array and name a table of their own, ten show a single column definition
and say (excerpt), three assign to $GLOBALS['TCA']['tt_content'] and name
an Overrides file, one is a user function class.

Releases: main, 14.3, 13.4
Signed-off-by: Lina Wolf


froemken (approved):
Adding captions to these `literalinclude` blocks provides huge value by making it instantly obvious whether a snippet represents a complete TCA file, a partial column configuration `(excerpt)`, a TCA override for `tt_content`, or a PHP class.

All 18 captions are syntactically clean (indented with 4 spaces), follow the official `EXT:my_extension/` naming scheme, and the PR targets `main` with appropriate backport labels.

PR #1449, opened 2026-09-12
[TASK] Include the central messages file instead of printing it

linawolf:
One page out of 91 uses literalinclude for /Includes.rst.txt where all
the others use include. The file holds nothing but a comment today, so
nothing is visible either way — but the moment a central message is put
there, every page shows it except this one, which would print the file as
code instead.

Releases: main, 14.3, 13.4
Signed-off-by: Lina Wolf


froemken (approved):
Good catch – /Includes.rst.txt must be included via `..  include::` to provide global definitions rather than being printed as code via `literalinclude`. Adheres to the canonical header sequence and reST standards.

PR #1466, opened 2026-09-13
[TASK] Fix the two reST indents #1463 missed

linawolf:
Two directive bodies sit one space off: the versionadded note in
Group/Index.rst indents five, and a nested definition in
FlexformSyntax.rst indents nine where its parent is at four.

#1463 converted this repository by hand from a list of files that a
too-narrow search had produced — it looked for three spaces at the start
of a line and therefore saw neither of these. A tool that walks the
constructs instead of the column finds them, and finds nothing else here
afterwards.

The rendered output is unchanged. All 93 pages were rendered before and
after and compared byte for byte.

Releases: main, 14.3, 13.4
Signed-off-by: Lina Wolf


linawolf:
The 13.4 backport had nothing to apply: #1465 redid this conversion
against 13.4's own files and already covered both of them, so the branch
carries the change. 14.3 arrives through #1467. Marking this done rather
than failed.

PR #1485, opened 2026-09-15
[TASK] Rename OuterWrapContainer to FormWrapContainer

linawolf:
TYPO3 v15 removed two FormEngine classes this manual still described:

`OuterWrapContainer` and its render type `outerWrapContainer` are replaced
by `FormWrapContainer` and `formWrapContainer`. The chapter on ctrl
container, its example and the registered node in the TCA snippet now name
the class that exists. The sentence saying the wizard HTML appears
"between the record title and the field body" went with it — record
headers are the controller's job now, so the wizard is rendered above the
fields.

The `tcaDescription` field information render type is gone as well. The
page keeps its anchors, since those stay once published, but its notice
says removed instead of deprecated and points at the breaking change.

References: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1748
Releases: main
Signed-off-by: Lina Wolf


sarahmccarthy123 (approved):
Looks good :-)

